PR #10, opened 2019-09-30
Enable AWS old releases

mazzy89:
Towards https://github.com/giantswarm/giantswarm/issues/6902

Enable AWS old releases

- 8.2.0
- 8.2.1
- 8.3.0

@pipo02mix I've enabled those past ones. Would make sense to enable more?

### DIFF
--- a/aws.yaml
+++ b/aws.yaml
@@ -78,7 +78,7 @@
       version: 0.19.0
   date: 2019-08-19T16:30:00Z
   version: 8.4.0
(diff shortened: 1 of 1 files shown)


pipo02mix:
I think it is ok 
thx

mazzy89:
Considering that there are other customers running also at 8.2.0 and they do want to create clusters at that version I'm going to consider to enable also for that version.

PR #11, opened 2019-09-30
Enable old KVM releases

mazzy89:
Towards https://github.com/giantswarm/giantswarm/issues/6902

Enabling 

- 8.2.0
- 8.2.1
- 8.3.0

8.2.0 -> 8.2.1: the unique difference is in the k8s version.

8.2.0 has a version affected by a security issue but it is not broken.


mazzy89:
Closing this becase we won't tackle it down in this way

PR #17, opened 2019-10-22
AWS release 9.0.0 with Kubernetes 1.15.5

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/6431

### DIFF
--- a/aws.yaml
+++ b/aws.yaml
@@ -1,3 +1,23 @@
+- active: false
+  authorities:
+    - endpoint: http://app-operator:8000
+      name: app-operator
+      version: 1.0.0
+    - endpoint: http://aws-operator:8000
+      name: aws-operator
+      version: 5.5.0
+    - endpoint: http://cert-operator:8000
+      name: cert-operator
+      version: 0.1.0
+    - endpoint: http://chart-operator:8000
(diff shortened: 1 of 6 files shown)


tfussell:
Should 8.X remain active? What about 7.X? Not sure about the deprecation process.

pipo02mix:
We only maintain the last two majors, so `7.X` goes deprecated

aws.yaml
```diff
@@ -1,4 +1,4 @@
-- active: false
+- active: true
```
kopiczko: aws-operator@5.5.0 is a legacy branch so we'd need to release a flat operator. Please ping @giantswarm/team-magic  when doing so for a while.

@giantswarm/team-magic needs to fix one bug with finalizers before we can deploy that to production. It was discovered yesterday evening. I'll link to the issue once it's created.

kopiczko: Here's the issue https://github.com/giantswarm/giantswarm/issues/7376

kopiczko: Fix is merged and vendored now..

MarcelMue:
I created a separate PR for KVM to make our lives a bit easier.

PR #18, opened 2019-10-25
remove endpoints

TheoBrigitte:
Towards: https://github.com/giantswarm/giantswarm/issues/7398

This PR removes the endpoints from all releases.

Since endpoint value from authorities is no longer used, and endpoint are now discovered automatically using authorities name.

```
$ for f in *.yaml; do yq -y -r 'del(.[].authorities[].endpoint)' $f|sponge $f; done
$ sed  -i "s/date:\s'\([0-9T:-]\+\)'/date: \1Z/g" *.yaml
```

### DIFF
--- a/aws.yaml
+++ b/aws.yaml
@@ -1,701 +1,537 @@
 - active: false
   authorities:
-    - endpoint: http://app-operator:8000
-      name: app-operator
-      version: 1.0.0
-    - endpoint: http://aws-operator:8000
-      name: aws-operator
-      version: 5.5.0
-    - endpoint: http://cert-operator:8000
-      name: cert-operator
-      version: 0.1.0
-    - endpoint: http://chart-operator:8000
-      name: chart-operator
-      version: 0.7.0
-    - endpoint: http://cluster-operator:8000
-      name: cluster-operator
-      provider: aws
-      version: 0.21.0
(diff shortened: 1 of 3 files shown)


xh3b4sd:
The build fails and I reckon there is a bit more to it. We need to fix the release validation which likely is an `opsctl` command. And we should make sure that the endpoint is then also removed from any code we have.

aws.yaml
```diff
@@ -1,701 +1,537 @@
...
+    version: 1.0.0
+  - name: aws-operator
+    version: 5.5.0
+  - name: cert-operator
+    version: 0.1.0
+  - name: chart-operator
+    version: 0.7.0
+  - name: cluster-operator
+    provider: aws
+    version: 0.21.0
```
kopiczko: This is personal taste thing but it's easier for me to read when lists are indented in yaml so if you don't mind I would rather have them indented like they were.

xh3b4sd: We should also split PRs for this. Dropping endpoints is one PR. Changing formatting is another PR. This makes the process and bringing changes forth way easier.

TheoBrigitte: I didn't intended to change the indentation, this was the result of the command. Closing here, and will re-work this later on.

PR #20, opened 2019-10-28
Update WIP releases to use cluster-operator 0.22.0

rossf7:
Updates the WIP releases to use the new version wired in https://github.com/giantswarm/cluster-operator/pull/752.

@MarcelMue @tfussell Can we use this for the new WIP releases you're preparing?

So we ship cluster-operator 0.21.0 in the new active versions.

### DIFF
--- a/aws.yaml
+++ b/aws.yaml
@@ -15,7 +15,7 @@
     - endpoint: http://cluster-operator:8000
       name: cluster-operator
       provider: aws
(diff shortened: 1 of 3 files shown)


tfussell:
Certainly. I'll update the releases before they're merged to match this.

rossf7:
Thanks @tfussell going with this.

PR #23, opened 2019-10-31
add Node Pools Pre-Release

xh3b4sd:
We deploy this to `seal` to give customers early access to our new Node Pools feature.


xh3b4sd:
Getting `9.x.x` in and have 3 active versions is fine IMO but that should be up to the SE maybe.

xh3b4sd:
stale

PR #29, opened 2019-11-07
Deactivate 8.4.1 on azure

T-Kukawka:


tfussell:
@T-Kukawka With your permission, I would like to close this PR in favor of https://github.com/giantswarm/releases/pull/38 which also disables 8.4.1 on Azure and releases 8.6.0 with the VMSS fix for all three providers.

PR #31, opened 2019-11-07
Change KVM and Azure to markdown

MarcelMue:


MarcelMue:
> When this is the desired formatting, is our tooling aligned with that or do we need different output for different "channels"?

The idea is for now to have this here as the main output. I want this to be atleast readable for tomorrow. 

This was discussed in sig-content but I need to make a proper announcement:
- We want to have proper releasenotes here in this repo
- We only want to send a small heads up with the release in Slack which links to this repo

PR #54, opened 2019-12-27
Add AWS 10.2.0 WIP release

sslavic:
Towards https://github.com/giantswarm/giantswarm/issues/6817

Supports testing cluster-operator master branch HEAD 2.1.0-dev version (see [here](https://github.com/giantswarm/cluster-operator/blob/master/pkg/project/project.go#L4)).

### DIFF
--- a/aws.yaml
+++ b/aws.yaml
@@ -1,3 +1,18 @@
+- active: false
+  authorities:
+    - name: app-operator
+      version: 1.0.0
+    - name: aws-operator
+      version: 8.0.2
+    - name: cert-operator
+      version: 0.1.0
+    - name: chart-operator
+      version: 0.7.0
+    - name: cluster-operator
+      provider: aws
+      version: 2.1.0-dev
+  date: 2019-12-27T14:00:00Z
+  version: 10.2.0
 - active: true
   authorities:
     - name: app-operator


aws.yaml
```diff
@@ -1,3 +1,18 @@
...
+      version: 1.0.0
+    - name: aws-operator
+      version: 8.0.2
+    - name: cert-operator
+      version: 0.1.0
+    - name: chart-operator
+      version: 0.7.0
+    - name: cluster-operator
+      provider: aws
+      version: 2.1.0-dev
```
kopiczko: I think we should bump bump the patch and not minor part of the version for components and releases. But I understand you didn't do it in the cluster-operator repo.

sslavic: Thanks for hint that this should be tracked as cluster-operator patch level change. What about GS release version?

https://github.com/giantswarm/giantswarm/issues/8135 is not there yet, so unfortunately we still have to bump cluster-operator for app changes.

More problematic is also whole GS release process. I'm still puzzled when should cluster-operator be released, and when GS release, who takes care of which step; asking around for guidance (see https://github.com/giantswarm/giantswarm/issues/6817#issuecomment-569639576 and discussions in #team-batman)

So even some technical steps could be added to https://github.com/giantswarm/giantswarm/blob/master/areas/empowerment/versioning-of-operators/how-to-release-an-operator.md but also (link to any existing?) product release process steps.

Once clarified, will get back to this and adjust version accordingly.

kopiczko: This is WIP release so you should put there whatever you find necessary to include in next release. I only wonder how we should deal with release changelogs cc @giantswarm/team-ludacris.

tfussell: You can see our general approach in this issue https://github.com/giantswarm/roadmap/issues/19 There are still some details to work out now that we have VOO and apps. Basically, we take the current HEAD of each component at a point in time and call that the release. We try to make sure active work gets merged in or is targeted to the next release. The important changelogs come from cluster-operator and the provider operator. Component updates in cluster-operator should have a corresponding note in the changelog (like https://github.com/giantswarm/cluster-operator/pull/860/files).

Ideally, the person that makes a change to an operator, whether adding, deleting, updating, etc. should also add it to the component changelog. I also try to look for code changes since the last release to make sure the changelog is correct, but it's error prone. We also keep customer-facing release notes in the releases repo now. It would be great if additions to the changelogs could also get added to the WIP release notes there (e.g. https://github.com/giantswarm/releases/blob/271d26679a083a63cc7c77e2e97d1bdf73b922c1/release-notes/aws/v11.0.0.md).

As far as semver for GS releases, we usually say a minor Kubernetes update is a major GS version (9.0->10.0), a patch Kubernetes update is a minor GS version (9.0->9.1), and changes to operators or component versions could be major (e.g. node pools), minor (e.g. load-balancer changes), or patch (e.g. CVE fixes). A better decision process could be helpful here.

Regarding timing, we plan to adopt something like the Kubernetes release process where we release on a regular basis using a code freeze and QA, but we haven't worked out the details yet. For now, we release as soon as we check all of the boxes in the release issue (healthy clusters, changelogs, release notes, upgrade tests, conformance tests, etc.).

Let me know if there are still questions about changelogs or releases and I'll try to clarify.

sslavic:
Release served its purpose for testing removal of CPU limits. Releases HEAD advanced in the meantime. CPU limits removal currently seems not urgent enough and is likely to be released with some more changes. Closing this PR.

PR #57, opened 2020-01-03
Release Giant Swarm for Azure 8.6.0

omegas27:
Towards https://github.com/giantswarm/giantswarm/issues/7302

### DIFF
--- a/azure.yaml
+++ b/azure.yaml
@@ -33,6 +33,21 @@
     - name: app-operator
       version: 1.0.0
     - name: azure-operator
+      version: 2.6.2
+    - name: cert-operator
+      version: 0.1.0
+    - name: chart-operator
+      version: 0.7.0
+    - name: cluster-operator
+      provider: azure
+      version: 0.20.0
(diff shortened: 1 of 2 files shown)


azure.yaml
```diff
@@ -29,6 +29,21 @@
...
+    - name: cert-operator
+      version: 0.1.0
+    - name: chart-operator
+      version: 0.7.0
+    - name: cluster-operator
+      provider: azure
+      version: 0.20.0
+  date: 2019-12-31T10:00:00Z
+  version: 8.6.0
+- active: false
```
MarcelMue: This should be true, right?

MarcelMue: Wait no - I am dumb ignore me.

omegas27: It is version 8.4.0. Do we want it active as 9.0.0 and 8.6.0 are active?

omegas27: It happens :)

omegas27:
@giantswarm/team-ludacris  should we close that one?

PR #68, opened 2020-01-10
Add new WIP releases

rossf7:
- Adds 10.1.2 for Node Pools with new cluster-operator 2.0.2-dev.
- Adds 9.1.1 for legacy with new cluster-operator 0.21.2-dev.

### DIFF
--- a/aws.yaml
+++ b/aws.yaml
@@ -12,7 +12,7 @@
       provider: aws
       version: 2.0.2-dev
   date: 2020-01-10T12:00:00Z
-  version: 10.1.1
+  version: 10.1.2
 - active: true
   authorities:
     - name: app-operator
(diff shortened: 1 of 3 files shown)


rossf7:
@kopiczko Thanks for checking on this.

aws.yaml
```diff
@@ -1,3 +1,18 @@
+- active: false
+  authorities:
+    - name: app-operator
+      version: 1.0.0
+    - name: aws-operator
+      version: 8.0.2
```
kopiczko: I think it should be 8.0.3-dev https://github.com/giantswarm/aws-operator/blob/4f75376aecaeb4392f944731883ef2ee8925cf25/pkg/project/project.go#L4

rossf7: Thanks I missed that. https://github.com/giantswarm/releases/pull/69 will fix it.

PR #70, opened 2020-01-13
WIP: Release apps on release repo

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/8135

It will add app release information into this repo so we don't need to change the hard-coded version in cluster-operator.


aws.yaml
```diff
@@ -1,4 +1,101 @@
...
+          type: Changed
+          urls:
+            - https://github.com/giantswarm/cert-exporter/blob/master/CHANGELOG.md#121-2019-12-24
+    - app: kube-state-metrics
+      version: 1.0.0
+      componentVersion: 1.9.0
+      use-upgrade-force: true
+      changelogs:
+        - description: Upgraded to kube-state-metrics new release 1.9.0
+          type: Changed
```
tomahawk28: I kept this type as it is but honestly, I don't think this is necessary since the description is already there, and most of them are Changed.

tomahawk28:
After the agreement upon the template itself, I will add azure, kvm by following commits.

aws.yaml
```diff
@@ -1,4 +1,101 @@
 - active: false
+  apps:
+    - app: chart-operator
+      version: 0.11.2
+      use-upgrade-force: true
+      changelogs:
```
MarcelMue: Not very happy with this - we need to figure out where changelogs are supposed to live. This will blow up these yamls massively.

rossf7:
@tomahawk28 IMO this should have been a draft PR. 

We can't merge this until the related cluster-operator dev has been done. Please make it clear that we're not ready to merge this yet.

I'll review in detail later.

tomahawk28:
It was a draft but I moved to ready status, sorry. Not sure how to move it back.

aws.yaml
tomahawk28: That's true. Changelogs would be messy and not sure it should be together with version information.

aws.yaml
rossf7: > we need to figure out where changelogs are supposed to live

@MarcelMue Yes I think it needs to be this repo but that we need to find a better home for the changelogs.

aws.yaml
```diff
@@ -1,4 +1,101 @@
 - active: false
+  apps:
+    - app: chart-operator
+      version: 0.11.2
+      use-upgrade-force: true
```
rossf7: This is configuration that shouldn't be in the releases repo. We use upgrade force for all apps except coredns and nginx-ingress-controller. It isn't safe for these apps as a forced upgrade could cause downtime.

aws.yaml
```diff
@@ -1,4 +1,101 @@
...
+      changelogs:
+        - description: Remove CPU Limits
+          type: Changed
+          urls:
+            - https://github.com/giantswarm/chart-operator/pull/335
+    - app: coredns
+      version: 1.1.3
+      componentVersion: 1.6.5
```
rossf7:
```suggestion
      componentVersion: 1.6.5
      version: 1.1.3
```

I think we should order the fields.

aws.yaml
tomahawk28: I would delete `use-upgrade-force` and put them as a same way we would put `namespace` information inside `/pkg/`.

aws.yaml
tomahawk28: @MarcelMue @rossf7 How about keeping release tracks on CHANGELOG.md for every installation type? Just as what we do in normal apps repo.

aws.yaml
rossf7: Yes I think having this as separate metadata is better. The releases YAML should just be about the components and versions.

I think cluster-operator/pkg is a good home but it could also be a apps.yaml file in this repo. As we want to keep the cluster-operator changes to a minimum. WDYT?

aws.yaml
rossf7: But how do we link the apps here with the changelogs in app repos?

We need to keep the versions and changelogs exposed by the version endpoint in cluster-operator.

```
$ kg port-forward cluster-operator-2-0-1-5486684645-mwv8x 8000:8000
$ curl http://localhost:8000/ | jq .
```

What about using this structure but in a separate `/changelogs/aws.yaml` file?

aws.yaml
rossf7: We need to keep `type` its part of the changelog format and we do use different values. e.g when adding new apps.

https://keepachangelog.com/en/1.0.0/

aws.yaml
rossf7: On 2nd thoughts I don't like the apps.yaml idea. I don't think apps metadata should be in this repo. 

@tomahawk28 How about starting with `cluster-operator/pkg/apps`?

The priority is to make updating the versions easier. The apps metadata changes much less often.

aws.yaml
tomahawk28: @rossf7 Only changelogs you mean by that, right?

rossf7: Yes just the changelogs so this file is cleaner. But I think you'll need the app and version in both files so the data can be merged.

tomahawk28: One thing that I don't like about a separate `/changelog/` files; it's not clear which changelogs related to which component version. That would be too trivial. Someone needs to write a new `release-note` by summarising all changelogs on it. 
At the moment, I can't think which way is better, any ideas?

tomahawk28: > But I think you'll need the app and version in both files so the data can be merged.

Yes, with this information, someone can merge them and write new `release-note` if only we keep changelogs as latest always.

rossf7: The release notes are written for humans whereas these changelog yamls will be for machines. So there will be some differences. But the yaml files should help with writing the release notes.

I think we will need a full history of the changelogs so yes the files will become big. We could have a file per release but AIUI the release yamls are mounted in a configmap so they can be accessed by cluster-operator.

What about updating the PR with the new structure so we can see what it looks like?

changelogs/aws.yaml
```diff
@@ -0,0 +1,75 @@
+- changelogs:
+  - component: chart-operator
```
tomahawk28: To order all changelogs based on `component`, I put the component name as the first key in this array.

rossf7: I think the file will get big so it will be easier to navigate if the changelogs are grouped by release. Something like this.

```
- release: 10.1.1
  - apps:
    - app: chart-operator
      - version: 0.11.2
        description: Remove CPU Limits.
        type: Changed
        urls:
          - https://github.com/giantswarm/chart-operator/pull/335
    - app: coredns
      - version: 1.1.3
        componentVersion: 1.6.5
        description: Fix HPA manifests for Kubernetes 1.16.
        type: Changed
        urls:
          - https://github.com/giantswarm/coredns-app/blob/master/CHANGELOG.md#v113-2020-01-08
```

WDYT?

pipo02mix: big +1

kopiczko: Is this a valid yaml? I think the value can be either string or list.

```
- release: 10.1.1
  - apps:
    - app: chart-operator
```

kopiczko: I would just bring the version as a first key here as a convention. Sorting this alphabetically isn't really helpful.

tomahawk28: @kopiczko I'm not sure what the first key as you suggest, would you mind to give me an example?

kopiczko: @tomahawk28 sorry I confused things a little with the `release:` key being in alphabetical order. 

But still this is not a valid yaml:

```
  - release: 10.1.1
    - component: chart-operator
```

This is:

```
  - release-10.1.1:
    - component: chart-operator
```

rossf7: Yes sorry I did my suggestion too fast. I like this. It's very human readable but should also be easy to parse in the cluster-operator code.

kopiczko: Yeah this is a general problem with dynamic keys.

I think this is also valid:

```
- "10.1.1":
  - component: chart-operator
```

tomahawk28: Do we really need a double quote? I tried a few validators online but it doesn't complain about that.

tomahawk28:
https://github.com/giantswarm/releases/pull/70#discussion_r366255858 (Comment from @rossf7) 
if no objection, I would jump into cluster-operator changes :)

rossf7:
@tomahawk28 Yes great. Makes sense to start on the cluster-operator changes.

changelogs/aws.yaml
kopiczko: Maybe we don't.

tomahawk28:
Deployed this change into `gauss` for testing cluster-service.

changelogs/aws/v10.1.1.yaml
```diff
@@ -0,0 +1,75 @@
+- changelogs:
```
MarcelMue: I am not strictly against a file like this but having to put the correct version in the name is kind of difficult.

Additionally this file should be named `v10.1.2` to match the changes up top?

tomahawk28: I copied these changelogs from `10.1.1` so it should keep as it is I think. 
Why has the version name on the file is difficult? When it's wrong, you just need to fix it after that 😁 .

rossf7: The release notes also have the version in the filename so I think that should be OK. We could also check it exists with CI. A check that the app versions exist in the catalog would also be useful IMO.

But the filename being wrong in the example is confusing. So I think that should be fixed.

tomahawk28: @rossf7 Do you mean by changelogs/aws/v10.1.1.yaml? It's same changelogs from 10.1.1. Is there something wrong with versioning?

rossf7: In releases/aws.yaml you're adding `v10.1.2.yaml` but this file is named `v10.1.1.yaml`. I think it should also be named `v10.1.2.yaml`.

You can add a dummy changelog if you want. We can't actually merge this until the cluster-operator and cluster-service change are merged and we agreed with Ludacris we would wait until after the K8s 1.16 releases.

tomahawk28: I would add dummy 10.1.2.yaml for that. Thanks.

app/config/override.yaml
```diff
@@ -0,0 +1,12 @@
...
+# chart-operator must be installed first so the chart CRD is
+# created in the tenant cluster.
+chart-operator:
+  chart:     "chart-operator",
+  namespace: "giantswarm",
+# CoreDNS's Upgrade force is disabled to avoid affecting customer workloads.
+coredns:
+  useUpgradeForce: false,
+net-exporter:
+  chart: "net-exporter",
```
kopiczko: Do we need `net-exporter` and `cert-exporter` in here?

You can also get rid of trailing commas.

app/config/override.yaml
```diff
@@ -0,0 +1,12 @@
+cert-exporter:
+  chart: "cert-exporter",
+# chart-operator must be installed first so the chart CRD is
+# created in the tenant cluster.
+chart-operator:
+  chart:     "chart-operator",
```
kopiczko:
```suggestion
```

app/config/override.yaml
tomahawk28: Only these two apps don't have `%s-app` suffix in their chart-name. I also would like to change this, but since we need to think about the compatibility between older releases, I would like to keep `chart` in here.

kopiczko: Ack.

kvm.yaml
```diff
@@ -8,7 +30,7 @@
       version: 0.7.0
     - name: cluster-operator
       provider: kvm
-      version: 0.22.1-dev
+      version: 0.21.4-dev
```
kopiczko: Downgrade?

tomahawk28: Probably I was about to make it the same as a legacy version in https://github.com/giantswarm/cluster-operator/blob/397bbf0076dd67b80aaad8e734a7d1bafd307a8e/pkg/project/project.go#L4
during my last manual tests. 

But it's already up to 0.21.5-dev, I will update it when this PR is out of the draft.

kopiczko:
This looks good to me now. But I'm not a real stakeholder here.

app/config/override.yaml
```diff
@@ -0,0 +1,14 @@
...
+chart-operator:
+  chart:     "chart-operator"
+  namespace: "giantswarm"
+# CoreDNS's Upgrade force is disabled to avoid affecting customer workloads.
+coredns:
+  useUpgradeForce: false
+net-exporter:
+  chart: "net-exporter"
+nginx-ingress-controller:
+  useUpgradeForce: false
```
rossf7: There is a comment about affecting customer workloads for coredns. Its also needed here.

PR #78, opened 2020-01-29
Adding default & override config files

tomahawk28:
Toward giantswarm/giantswarm#8135

Same as https://github.com/giantswarm/releases/pull/70 but without changelogs.

For the spec, please check https://github.com/giantswarm/giantswarm/blob/master/areas/managed-services/specs/app-on-release.md


app/config/override.yaml
```diff
@@ -0,0 +1,15 @@
+cert-exporter:
+  chart: "cert-exporter"
+# chart-operator must be installed first so the chart CRD is
+# created in the tenant cluster.
+chart-operator:
+  chart:     "chart-operator"
+  namespace: "giantswarm"
+# CoreDNS's Upgrade force is disabled to avoid affecting customer workloads.
```
rossf7:
```suggestion
# Upgrade force is disabled to avoid affecting customer workloads.
```

app/config/override.yaml
```diff
@@ -0,0 +1,15 @@
...
+coredns:
+  useUpgradeForce: false
+net-exporter:
+  chart: "net-exporter"
+# nginx's Upgrade force is disabled to avoid affecting customer workloads.
```
rossf7:
```suggestion
# Upgrade force is disabled to avoid affecting customer workloads.
```

PR #81, opened 2020-01-29
Clean up 11.0 and 11.1 (WIP) releases

tfussell:
- Fix 11.0 and 11.1 release dates (copy and paste error)
- Add stub 11.1 release notes (trying this out for the next release)
- Remove cert-manager from release notes (only used in AWS w/ node pools AFAICT)

### DIFF
--- a/release-notes/azure/v11.1.0.md
+++ b/release-notes/azure/v11.1.0.md
@@ -1,0 +1,8 @@
+## :zap: Giant Swarm Release 11.1.0 for Azure is now active for you! :zap:
+
+_Introduction here_
+
+---
+
+### azure-operator v2.9.0
+- TODO.
--- a/release-notes/azure/v11.1.0.md
+++ b/release-notes/azure/v11.1.0.md
@@ -1,8 +1,0 @@
-## :zap: Giant Swarm Release 11.1.0 for Azure is now active for you! :zap:
-
-_Introduction here_
-
----
-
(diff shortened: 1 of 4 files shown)


release-notes/azure/v11.1.0.md
```diff
@@ -0,0 +1,8 @@
+## :zap: Giant Swarm Release 11.1.0 for Azure is now active for you! :zap:
```
MarcelMue: Should we have empty release notes in here? Kind of confusing IMO

MarcelMue: Just thinking of the angle of a customer looking at our repo.

tfussell: I thought it would be nice for us to have a running record of what has changed to avoid the work at the end. Maybe there's a better place for it though--I can take it out.

MarcelMue: Maybe adding a PR template could help?

tfussell: This template would just say to mention the change in the release notes? The question is whether in the future we expect to build up releases with multiple PRs or a single PR that is tested all together and then released. In the first case, a PR template would be good.

MarcelMue: Yes just having a checklist would be good IMO. I mean we can remove the boxes when they don't apply but it's just a good reminder I believe.

PR #92, opened 2020-02-05
App & apps' changelog on release 

tomahawk28:
Toward giantswarm/giantswarm#8135

App list on AWS only for node pool cluster. 
All changelogs, app versions are updated as the latest today (Feb 5). 

Spec: https://github.com/giantswarm/giantswarm/blob/master/areas/managed-services/specs/app-on-release.md

It will close https://github.com/giantswarm/releases/pull/70

### DIFF
--- a/aws.yaml
+++ b/aws.yaml
@@ -1,4 +1,4 @@
-- active: true
+- active: false
   apps:
     - app: chart-operator
       version: 0.11.4
@@ -19,8 +19,8 @@
       version: 1.2.0
       componentVersion: 0.18.1
     - app: cert-manager
-      version: 1.1.0
-      componentVersion: 0.13.0
+      version: 1.0.4
+      componentVersion: 0.9.0
     - app: cluster-autoscaler
       version: 1.1.4
       componentVersion: 1.16.2
@@ -30,6 +30,21 @@
     - app: kiam
       version: 1.1.0
       componentVersion: 3.5.0
+  authorities:
+    - name: app-operator
+      version: 1.0.0
+    - name: aws-operator
+      version: 8.1.0
+    - name: cert-operator
+      version: 0.1.0
+    - name: chart-operator
+      version: 0.7.0
+    - name: cluster-operator
+      provider: aws
+      version: 2.1.0-dev
+  date: 2020-02-05T12:00:00Z
+  version: 11.0.1
+- active: true
   authorities:
     - name: app-operator
       version: 1.0.0
--- a/aws.yaml
+++ b/aws.yaml
@@ -34,7 +34,7 @@
     - name: app-operator
       version: 1.0.0
     - name: aws-operator
-      version: 8.1.0
+      version: 8.1.1-dev
     - name: cert-operator
       version: 0.1.0
     - name: chart-operator
--- a/aws.yaml
+++ b/aws.yaml
@@ -3,33 +3,33 @@
     - app: chart-operator
       version: 0.11.4
     - app: coredns
-      version: 1.1.3
       componentVersion: 1.6.5
+      version: 1.1.3
     - app: cert-exporter
       version: 1.2.1
     - app: kube-state-metrics
-      version: 1.0.0
       componentVersion: 1.9.0
-    - app: metrics-server
       version: 1.0.0
+    - app: metrics-server
       componentVersion: 0.3.3
+      version: 1.0.0
     - app: net-exporter
       version: 1.5.1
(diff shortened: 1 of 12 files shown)


aws.yaml
```diff
@@ -1,4 +1,35 @@
 - active: true
+  apps:
```
rossf7: This is updating the 11.0.0 release but it's already active. Please add a 11.0.1 WIP release with the new cluster-operator release and apps.

This is easy to miss as the current master doesn't have a WIP for 11.X.

tomahawk28: Should be 11.1.0? since it's not bug fix actually.

rossf7: Let's use patch for now. These changes are about improving our developer experience. This is more interesting for us than for customers.

When this WIP version is released we can upgrade to a minor version if necessary.

aws.yaml
```diff
@@ -1,3 +1,49 @@
...
+    - app: net-exporter
+      version: 1.5.1
+    - app: node-exporter
+      version: 1.2.0
+      componentVersion: 0.18.1
+    - app: cert-manager
+      version: 1.0.4
+      componentVersion: 0.9.0
+    - app: cluster-autoscaler
+      version: 1.1.4
```
rossf7: There is no tarball for this release. https://github.com/giantswarm/default-catalog Can you check the version?

This is very easy to miss and I think we should have an automated check. So we know the versions here actually exist in the catalog.

@puja108 @piontec WDYT? Should we have an issue for that?

aws.yaml
```diff
@@ -1,3 +1,49 @@
...
+      version: 1.2.0
+      componentVersion: 0.5.18
+    - app: kiam
+      version: 1.1.0
+      componentVersion: 3.5.0
+  authorities:
+    - name: app-operator
+      version: 1.0.0
+    - name: aws-operator
+      version: 8.1.0
```
rossf7: This should also be WIP. https://github.com/giantswarm/aws-operator/blob/de013a3681525e3b47525a7e57989955c87ad226/pkg/project/project.go#L4

aws.yaml
```diff
@@ -1,3 +1,49 @@
...
+    - name: cert-operator
+      version: 0.1.0
+    - name: chart-operator
+      version: 0.7.0
```
rossf7: This chart-operator version is legacy cruft. It isn't the version used and I'd actually like to remove it.

@MarcelMue WDYT?

aws.yaml
```diff
@@ -1,3 +1,49 @@
+- active: false
+  apps:
+    - app: chart-operator
+      version: 0.11.4
+    - app: coredns
+      version: 1.1.3
+      componentVersion: 1.6.5
```
rossf7: Let's also order the version fields. I think that is tidier.

aws.yaml
MarcelMue: Works for me. No need to keep dead weight.

aws.yaml
tomahawk28: It was in a changelog but not released yet. I just publish a new 1.1.4 tag there, Thanks for the notice.

aws.yaml
tomahawk28: I put it like deliberately, but let's keep in order as you said.

aws.yaml
rossf7: Thanks I'd still like to have the automated check. But I think it can be done after these changes.

corest:
I've probably missed spec discussion, but maybe can get answer here - why do we go with `componentVersion` while charts actually use `appVersion`?

rossf7:
> I've probably missed spec discussion,

@tomahawk28 Can you link the spec in the PR desc?

> why do we go with componentVersion while charts actually use appVersion?

@corest In the API responses we have components and versions. So it aligns with that.

We also decided we wanted to be more opinionated on naming and not always align with Helm. 

For example in the next version of the app CRD `namespace` will probably be renamed to `targetNamespace`.

tomahawk28:
@corest `appVersion` is misleading the context a bit like it's like giantswarm's app version despite the fact that we have `version` for itself.  
Compare to `componentVersion`, it gives a sense that it's the version of the component inside giantswarm's app.

corest:
Got it. Thx for explanation @tomahawk28 @rossf7

tomahawk28:
going to ping @giantswarm/team-ludacris . 
Sorry

tomahawk28:
Going to merge.

PR #94, opened 2020-02-06
Use cluster-operator 0.22.2-dev for Azure WIP

rossf7:
I think this should be the same as KVM.


rossf7:
I don't think this is controversial.

Tested on `ghost`. Going with a ping.

PR #101, opened 2020-02-12
Put version, active, date as first 

tomahawk28:
As I mentioned this in #sig-releng channel, It would be much easier to look into the version in the first line of release rather than scrolling down to a file.

### DIFF
--- a/azure.yaml
+++ b/azure.yaml
@@ -1,4 +1,6 @@
-- active: false
+- version: 11.1.0
+  active: false
+  date: 2020-01-29 12:00:00+00:00
   apps:
     - app: cert-exporter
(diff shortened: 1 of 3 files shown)


azure.yaml
```diff
@@ -25,35 +27,35 @@
...
-  - name: azure-operator
-    version: 2.8.0
-  - name: cert-operator
-    version: 0.1.0
-  - name: chart-operator
-    version: 0.7.0
-  - name: cluster-operator
-    provider: azure
-    version: 0.21.4
+    - name: app-operator
```
tomahawk28: I found out these parts need an indentation. Not sure how it was working before.

PR #102, opened 2020-02-12
Put version, active, date first from release

tomahawk28:
As I mentioned this in #sig-releng channel, It would be much easier to look into the version in the first line of release rather than scrolling down to a file.

### DIFF
--- a/azure.yaml
+++ b/azure.yaml
@@ -1,6 +1,6 @@
 - version: 11.2.0
   active: false
-  date: 2020-01-29T12:00:00Z'
+  date: 2020-01-29T12:00:00Z
   apps:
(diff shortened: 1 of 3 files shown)


azure.yaml
```diff
@@ -25,20 +27,20 @@
...
-    version: 0.1.0
-  - name: chart-operator
-    version: 0.7.0
-  - name: cluster-operator
-    provider: azure
-    version: 0.23.0
-  date: 2020-01-29T12:00:00Z
-  version: 11.2.0
-- active: false
+    - name: app-operator
```
tomahawk28: I found out these parts need an indentation. Not sure how it was working before.

PR #105, opened 2020-02-19
Register cert-manager 1.0.5 and use it in AWS WIPs

sslavic:
Towards https://github.com/giantswarm/giantswarm/issues/8939


app/changelog/cert-manager.yaml
```diff
@@ -1,9 +1,9 @@
-- version: 1.1.0
```
sslavic: @corest this PR also removed cert-manager 1.1.0 app - I see you reverted that release via https://github.com/giantswarm/cert-manager-app/pull/12 in app repo but it was left over in this releases app registry.

PR #108, opened 2020-02-20
activate release 11.0.1

xh3b4sd:
This activates `11.0.1`, and deactivates `11.0.0` and `10.1.2`.


calvix:
should 11.0.0 be set to not active?

xh3b4sd:
Good point. Sorted.

xh3b4sd:
What about the active `10.x.x`release? I understood it is broken as well but this has the other Kubernetes version. Should it be deactivated as well?

teemow:
just check if there are clusters running with this release. you can talk to the responsible SE and ask if there is an upgrade planned. If there are no important clusters you can deactivate it. If there is a problem why customers are not upgrading you might need to consider a patch release for 10.x too.

xh3b4sd:
Checked with @giantswarm/se and deactivated `10.1.2`. We help customers move to `11.x.x` and in case a patch for `10.x.x` is necessary we deliver it separately. This does not seem to be necessary but we did not get feedback on all clusters. We are moving now.

PR #84, opened 2020-01-30
[WIP] Draft release notes for Azure v11.2.0

axbarsan:

PR #112, opened 2020-02-20
Define and activate AWS 9.2.0 legacy release

sslavic:


sslavic:
Upgrading from 9.1.0 to 9.2.0, and clean 9.2.0 install have been tested in giraffe (AWS, China)

sslavic:
Upgrading from 9.1.0 to 9.2.0, and clean 9.2.0 install have been tested on ginger (AWS, EU)

xh3b4sd:
For my interest, why is this a minor bump? Why not major or patch?

sslavic:
nginx default ssl ciphers have changed, dropping support for some older browsers; we've made appropriate note in release notes, still we want to bring attention to this user affecting change in additional way through making this minor bump.

aws.yaml
```diff
@@ -146,6 +146,44 @@
...
+      componentVersion: 0.29.0
+      version: 1.5.0
+    - app: node-exporter
+      componentVersion: 0.18.1
+      version: 1.2.0
+  authorities:
+    - name: app-operator
+      version: 1.0.0
+    - name: aws-operator
+      version: 5.6.0
```
pipo02mix: Why chart operator is not listed in this last one as authority?

sslavic: @rossf7 explained in https://github.com/giantswarm/releases/pull/84#discussion_r381200033

If I'm not mistaken, same rationale applies for legacy aws releases like this too.

PR #114, opened 2020-02-20
Define and activate Azure 11.2.0 release

sslavic:


sslavic:
I've tested on ghost:
- 11.2.0 clean cluster creation
- upgrade from 11.1.0 to 11.2.0

In both cases cluster would have batches of nods NotReady, sometimes even all worker nodes; @giantswarm/team-celestial claims this is known issue https://github.com/giantswarm/giantswarm/issues/9052

Also, it seems 11.1.0 ships with external-dns as default app, it's supposed to be used on aws node pool clusters only - I've shared finding with @giantswarm/team-celestial and ExternalName Service issue dvag reported to them might be related to this; it's a dilemma whether they want to release 11.1.1 as patch release to fix/remove external-dns, or 11.2.0 would be treated as fix, and might need to disable 11.1.0 release to prevent (actually just suggest to avoid) creation of new clusters with this known issue.

Besides these issues, nginx installation and update with 11.2.0 legacy WIP release seems to go well.

rossf7:
@sslavic On external-dns it is a default app on Azure and has been for a long time. This is the old chart from when it was deployed as a chartconfig CR. https://github.com/giantswarm/kubernetes-external-dns

Its kiam which is a new default app for AWS CAPI clusters and should not exist for legacy AWS.

sslavic:
Reintroduced external-dns and used updated verision external-dns-app-1.2.0 compared to 11.1.0. Will update azure 11.2.0 release notes accordingly.

Retested clean install and upgrade on ghost, observed https://github.com/giantswarm/giantswarm/issues/9052 again, besides that install and upgrade worked.

cokiengchiara:
@sslavic can we collaborate on release notes from Dropbox Paper? https://www.dropbox.com/scl/fi/28ed8z7mzwr9vit62p4xk/_-Azure-and-KVM-11.2.0.paper?dl=0&rlkey=4kpr3k1ctwfzos3iq9l0ig848

1. Let's not write it on any PR since we have to duplicate the content into two PRs
2. Testing DB Paper instead of HackMD for easier collaboration (can export to MD, easier commenting, better readability)

PR #119, opened 2020-02-21
Use kiam-app v1.1.1 in next AWS node pools WIP

sslavic:
kiam-app v1.1.1 includes internal improvement, globally configurable image registry for parent and subchart, in China installations to use aliyun registry instead of quay https://github.com/giantswarm/kiam-app/pull/15

Registry to be used is already configured in default app catalog, axolotl and giraffe specific overrides, and app catalogs ensured/updated.


sslavic:
@rossf7 I added changelog entry, PTAL

PR #122, opened 2020-02-21
Move KVM release notes 11.1.1 to 11.2.0

sslavic:
Due to breaking change in the next KVM WIP release, instead of bumping patch we're bumping minor version number. This PR renames release version in release notes accordingly.


sslavic:
Going with a ping

PR #125, opened 2020-02-24
Define and activate KVM 11.3.0 release

mazzy89:


kvm.yaml
```diff
@@ -1,3 +1,40 @@
...
+      version: 1.0.0
+    - name: cert-operator
+      version: 0.1.0
+    - name: cluster-operator
+      provider: kvm
+      version: 0.23.1
+    - name: flannel-operator
+      version: 0.2.0
+    - name: kvm-operator
+      version: 3.11.0
```
mazzy89: This should be right and the next tag we are suppose to create under kvm-operator

kvm.yaml
```diff
@@ -69,6 +106,23 @@
       version: 0.2.0
     - name: kvm-operator
       version: 3.10.0
+- version: 9.3.0
```
mazzy89: 9.3.0. there is no 9.1.0 and 9.2.0 but I recall @MarcelMue that you talked about 9.3.0. Am I wrong?

kvm.yaml
```diff
@@ -69,6 +106,23 @@
...
+      version: 0.1.0
+    - name: chart-operator
+      version: 0.7.0
+    - name: cluster-operator
+      provider: kvm
+      version: 0.21.0
+    - name: flannel-operator
+      version: 0.2.0
+    - name: kvm-operator
+      version: 3.9.0
```
mazzy89: This is tough. Which should be the value here?

MarcelMue: Keep the `-dev` version you had previously for now - I want to at least create a cluster and check some apps.

kvm.yaml
MarcelMue: I might have confused `kvm` and `aws` - `9.1.0` looks correct for kvm.

kvm.yaml
MarcelMue: Sorry the previous comment was meant to be on line 37.

Here we will need to do some more magic. I think this will require some changes in `thiccc` branch of kvm-operator. Let's talk tomorrow.

MarcelMue:
@giantswarm/team-ludacris We should test this manually together atleast once to check apps and so on.

sslavic:
@MarcelMue @mazzy89 what are the plans to release this release? If this week, we'd like to include nginx IC 1.6.7 App in it, to include the fix for graceful termination misconfiguration. Otherwise we'd kick off a new kvm 11.2.x patch release.
//cc @cokiengchiara

MarcelMue:
> @MarcelMue @mazzy89 what are the plans to release this release? If this week, we'd like to include nginx IC 1.6.7 App in it, to include the fix for graceful termination misconfiguration. Otherwise we'd kick off a new kvm 11.2.x patch release.
> //cc @cokiengchiara

We will need some more time here - please feel free to add a patch release instead.

MarcelMue:
Closing in favor of #241

PR #131, opened 2020-02-26
Remove "future releases" for kvm

cokiengchiara:


sslavic:
@cokiengchiara shall we merge this one?

PR #133, opened 2020-02-27
Update README with missed release

MarcelMue:


xh3b4sd:
I think I missed this part. Can we have some CI check for that? Should be doable to parse for that section and compare the markdown list with the release files in the folder.

sslavic:
@MarcelMue can you please merge this one?

PR #135, opened 2020-02-28
New app changelog entries

sslavic:
This PR adds changelog entries for:
- cert-manager app 1.0.6
- nginx IC app 1.6.0

so they can be referenced in releases (release definition/activation/notes PRs will follow).


sslavic:
Going with a ping

PR #136, opened 2020-02-28
Define and activate AWS 9.2.1 release

sslavic:
This also deactivates 9.2.0

PR should be approved only once release notes are prepared and ready to share.


sslavic:
Tested on gauss.

@cokiengchiara I'm ready to ship the release when you are ready to share the announcement.

PR #137, opened 2020-02-28
bump kube-state-metrics to v1.0.3

wtty-fool:
Towards https://github.com/giantswarm/giantswarm/issues/8569

### DIFF
--- a/app/changelog/kube-state-metrics.yaml
+++ b/app/changelog/kube-state-metrics.yaml
@@ -1,7 +1,7 @@
 - version: 1.0.3
   componentVersion: 1.9.2
   description: Remove CPU and memory limits.
-  kind: fixed
+  kind: changed
   urls:
     - https://github.com/giantswarm/kube-state-metrics-app/blob/master/CHANGELOG.md#v103
 - version: 1.0.2
--- a/app/changelog/kube-state-metrics.yaml
+++ b/app/changelog/kube-state-metrics.yaml
@@ -1,3 +1,9 @@
+- version: 1.0.3
+  componentVersion: 1.9.2
+  description: Remove CPU and memory limits.
+  kind: fixed
+  urls:
+    - https://github.com/giantswarm/kube-state-metrics-app/blob/master/CHANGELOG.md#v103
 - version: 1.0.2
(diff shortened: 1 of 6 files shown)


azure.yaml
```diff
@@ -14,7 +14,7 @@
       version: 1.2.0
     - app: kube-state-metrics
       componentVersion: 1.9.2
-      version: 1.0.2
+      version: 1.0.3
```
sslavic: IMO this shouldn't be done - changing the definition of already published release.

@wtty-fool would it make sense to keep this PR only for adding 1.0.3 kube-state-metrics to the apps changelog? We're in the process of releasing next patch releases (see other PRs in the repo), would be happy to include kube-state-metrics v1.0.3 in them

kvm.yaml
```diff
@@ -11,7 +11,7 @@
       version: 1.1.3
     - app: kube-state-metrics
       componentVersion: 1.9.2
-      version: 1.0.2
+      version: 1.0.3
```
sslavic: This is also published release, they shouldn't be modified.

azure.yaml
kopiczko: Thanks @sslavic for spotting that. Then we can include that only for AWS as this is apparently the only provider with WIP release in place... Is that ok?

sslavic: Sure, once you merge I'll take it into separate already created PR which activates that release and makes more changes to the release definition/package to ship them all together as one.

wtty-fool: @sslavic Good catch, thank you for pointing that out.
Ok, I'll keep an update to AWS and changelog, the rest will be dropped from this PR.

app/changelog/kube-state-metrics.yaml
```diff
@@ -1,3 +1,9 @@
+- version: 1.0.3
+  componentVersion: 1.9.2
+  description: Remove CPU and memory limits.
+  kind: fixed
```
sslavic:
```suggestion
  kind: changed
```

sslavic: If for nothing else, then for consistency with equivalent changelog entries done for other apps before.

PR #139, opened 2020-02-28
Define KVM 11.2.1 WIP release

sslavic:
This also deactivates 11.1.0, which should have been deactivated with 11.2.0 release.

### DIFF
--- a/kvm.yaml
+++ b/kvm.yaml
@@ -19,7 +19,7 @@
       version: 1.6.0
     - app: nginx-ingress-controller
       componentVersion: 0.30.0
-      version: 1.6.4
+      version: 1.6.5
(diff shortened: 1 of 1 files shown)


sslavic:
@giantswarm/chapter-se would it be acceptable to ship this release with the known issue that nginx will not start if all that tenant cluster has
- 2CPU worker nodes
- single worker node with 4 or less CPU?

rossf7:
Tested upgrade from 9.0.0 on `geckon`. All good.

sslavic:
@cokiengchiara I'll merge this WIP release definition to be able to more easily finish final tests. I'll create a separate PR for release activation/publishing to customers.

PR #144, opened 2020-03-04
Bump kube-state-metrics to v1.0.4

wtty-fool:


wtty-fool:
Tested on `gauss`. Merging.

PR #142, opened 2020-03-02
circleci: Push to app collection

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/8655 and https://github.com/giantswarm/giantswarm/issues/8436

This PR implements a CircleCI workflow for pushing releases as a helm check to the catalog and provider app collections so they can be automatically deployed to control planes.

PR #147, opened 2020-03-09
Prepare release definitions for Release CRs

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/8436 and split off from https://github.com/giantswarm/releases/pull/142

We need certain information in Release CR that isn't currently present in our `aws/azure/kvm.yaml` release definitions. This PR adds/transforms them in a backwards compatible way:
- WIP releases have `state: wip` added.
- Deprecated releases have `state: deprecated` added.
- Active releases have `state: active` added.
- Each release now has a list of `components` copied from the provider operator version bundle (very tedious). This gives us effective parity between `gsctl list releases` and `kubectl get releases` even for older releases.

`components` and `state` should be ignored by draughtsman/cluster-service so I don't think this will affect our current release process.


tfussell:
> LGTM on a quick scan - would need to double triple check when we actually use these, that no version changes.

I just double checked again by comparing against `gsctl list releases` and found one error. I will probably need another set of eyes on it before I feel comfortable merging though.

tfussell:
I'm going to go ahead and merge this to keep things moving fast since it's already approvedd. Let's just be sure to confirm versions before we actually use releases.

PR #148, opened 2020-03-09
circleci: Push release CRs to control planes

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/8655 and https://github.com/giantswarm/giantswarm/issues/8436 and split off from https://github.com/giantswarm/releases/pull/142

This PR implements a CircleCI workflow for pushing releases as a helm chart to the control plane catalogs and provider app collections so they can be automatically deployed to control planes.

Here's the provider-specific commits on the `control-plane-test-catalog`: [aws](https://github.com/giantswarm/control-plane-test-catalog/commit/1e2ba3b9cdd9619cd21c4591bd3049763bd70b65) [azure](https://github.com/giantswarm/control-plane-test-catalog/commit/9ee27813afad73a7de8485ea43f93ac2a08643b1) [kvm](https://github.com/giantswarm/control-plane-test-catalog/commit/7f6c47c3acce75ade273e4381a358f9fc8358b65) from the latest build. I installed the chart on `ginger` manually using `helm install https://giantswarm.github.io/control-plane-test-catalog/releases-kvm-0.0.0-a0e045ed25f471017f324f12d19322d6a45236d1.tgz --name releases-kvm-unique`. The output of `kubectl get releases` is as follows:
```
kubectl get releases
NAME      KUBERNETES VERSION   STATE        AGE
v11.0.0   1.16.3               deprecated   56d
v11.1.0   1.16.3               active       37d
v11.2.0   1.16.3               active       9d
v8.0.0    1.14.1               deprecated   351d
v8.1.0    1.14.1               deprecated   311d
v8.2.0    1.14.3               deprecated   256d
v8.2.1    1.14.4               deprecated   256d
v8.3.0    1.14.5               deprecated   199d
v8.4.0    1.14.6               deprecated   199d
v9.0.0    1.15.5               active       130d
```
On merge to `master` this will go out on all control planes which I think is fine. Future PRs will add support for reading and using Release CRs to determine component images.


.circleci/config.yml
```diff
@@ -1,17 +1,145 @@
...
+          source: aws
+      - architect/push-to-app-catalog:
+          name: push-releases-aws-to-control-plane-app-catalog
+          app_catalog: control-plane-catalog
+          app_catalog_test: control-plane-test-catalog
+          attach_workspace: true
+          chart: releases-aws
+          requires:
+            - prepare-helm-chart-aws
+            - push-releases-kvm-to-control-plane-app-catalog
```
tfussell: This looks weird, but it's required to avoid a race when multiple `architect/push-to-app-catalog` jobs are committing to the same repo. They need to run in sequence.

.circleci/release.yaml
```diff
@@ -0,0 +1,33 @@
+{{- range .Values.versions }}
+apiVersion: release.giantswarm.io/v1alpha1
+kind: Release
+metadata:
+  annotations:
+    giantswarm.io/docs: https://pkg.go.dev/github.com/giantswarm/apiextensions/pkg/apis/release/v1alpha1?tab=doc#Release
```
tfussell: This should point to our new docs website once it's live.

azure.yaml
```diff
@@ -41,7 +41,7 @@
     - name: kubernetes
       version: 1.16.3
     - name: containerlinux
-      version: 2247.6.0
+      version: 2191.5.0
```
tfussell: I made a mistake on the 1.16.3 release for Azure. I changed the `containerlinux` version bundle but didn't actually change the version so I'm reverting it here to match reality.

.circleci/release.yaml
MarcelMue: Do we have an issue tracking this or something?

.circleci/config.yml
MarcelMue: It actually makes sense to me - just the expected circleci dance.

.circleci/release.yaml
tfussell: I guess it's this one https://github.com/giantswarm/giantswarm/issues/8804

kopiczko:
Is there a plan to store plain CRs instead? This CI workflow is pretty magical.

MarcelMue:
Yes - this is an intermediary step while all our components still use the yaml. It prevents us from diverging between yaml and CRs.
Once we switch to using the CRs in our components, then we will remove the CI and instead have CRs directly in this repo.

PR #163, opened 2020-03-17
circleci: Fix catalog name for collection job

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/8436 and related to https://github.com/giantswarm/installations/pull/1116

For `architect/push-to-app-catalog`, `app_catalog` refers to the name of the repository to push to but in `architect/push-to-app-collection` it refers to name of the `appcatalog` CR on the CP. In this case, these values are `releases-catalog` and `releases` respectively so `architect/push-to-app-collection` must be changed to reflect this difference.

While testing I noticed that newer releases (those with `apps`) have `coredns` defined in two places so I'm taking this opportunity to remove them from `components` as well.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -83,7 +83,7 @@ workflows:
             - prepare-helm-chart-kvm
       - architect/push-to-app-collection:
           name: push-releases-to-kvm-app-collection
-          app_catalog: releases-catalog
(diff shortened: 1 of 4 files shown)


tfussell:
App catalogs have been ensured on all installations so I will merge this after CI completes.

PR #164, opened 2020-03-17
Deactivate release 11.1.1

paurosello:
Deactivate release 11.1.1 until further notice


sslavic:
If activating 11.1.1 was mistake, that it will take day or so to actually release it, maybe merge deactivation as soon as possible to minimize chance of someone upgrading to it while it's WIP, go with the notification don't wait for review to merge.

PR #166, opened 2020-03-17
Cover app updates in AWS 11.1.1 RNs

sslavic:
I've taken these release notes from canceled 11.0.2 WIP

### DIFF
--- a/aws.yaml
+++ b/aws.yaml
@@ -7,7 +7,7 @@
       version: 1.2.1
     - app: cert-manager
       componentVersion: 0.9.0
-      version: 1.0.5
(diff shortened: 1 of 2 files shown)


aws.yaml
```diff
@@ -7,7 +7,7 @@
       version: 1.2.1
     - app: cert-manager
       componentVersion: 0.9.0
-      version: 1.0.5
+      version: 1.0.6
```
sslavic: @paurosello I've bumped cert-manager too in this PR, taken over from 11.0.2 release definition.

paurosello: thanks!

PR #174, opened 2020-03-18
Use latest aws-operator 8.2.2-dev in wip version 11.1.1

corest:


calvix:
isnt 11.1.1 used for something @paurosello ?

paurosello:
Yes, there are 2 cluster in IC Consult.

We are going to need a new version with the updated net-exporter version.
Preferably without changing the AWS operator so we don't have to roll
nodes...


What do you think?

corest:
It is in WIP, so I'm updating it. If it is not a wip anymore, why isn't it active?

paurosello:
We don't want it to be active in all customers, can't we have another WIP
version?

corest:
> We don't want it to be active in all customers, can't we have another WIP version?

corest:
> We don't want it to be active in all customers, can't we have another WIP version?

We probably can. But then releasing team needs to add new wip release and announce that `11.1.1` should stay untouched and unreleased. Release message here https://gigantic.slack.com/archives/C90BZMUJ1/p1584440819128100 doesn't mention this. Also after adding new wip and then releasing it - previous one will become deprecated and visible for all the other customers

paurosello:
Then could we at least change this version to 11.1.2?

IC Consult has two clusters in 11.1.0 and two with 11.1.1 and I will be upgrading them. No need to add 11.1.0 and 11.1.1 to master, will create a branch to fix all the mess.

PR #186, opened 2020-03-23
Register NGINX IC App v1.6.5

sslavic:


sslavic:
Non-controversial change, going with a ping

PR #201, opened 2020-03-31
Add AWS release v11.1.2

marians:
Internal issue: https://github.com/giantswarm/giantswarm/issues/9766


aws.yaml
```diff
@@ -1,6 +1,6 @@
-- version: 11.1.1
+- version: 11.2.0
```
xh3b4sd: Is it intended to change this version instead of adding a new release? Looks wrong to me now, also because the issue linked in the description talks about `aws-operator@8.2.2` while the operator version here did not change. Do we really have everything in the set tag?

marians: The aws-operator release is not done yet. But I guess we can already set it here.

There is no more need for the inactive 11.1.1 release, so we are effectively replacing it.

release-notes/aws/v11.2.0.md
```diff
@@ -1,8 +1,16 @@
-# :zap: Giant Swarm Release 11.1.1 for AWS is now active for you! :zap:
```
MarcelMue, line 1: Why does the title completely break convention now?

release-notes/aws/v11.2.0.md
```diff
@@ -1,8 +1,16 @@
...
+# AWS v11.2.0
 
-## aws-operator 8.2.1
+__Note:__ Upgrading to this release from any release prior 11.1.1 will cause a network downtime due to the network-related changes coming with the switch from Calico CNI to AWS CNI.
 
-- Replaces Calico CNI (Container Network Interface) with [AWS CNI](https://github.com/aws/amazon-vpc-cni-k8s) and enables Pod to Pod networking throughout AWS regions. As a result of this change, CIDR (IPv4 address blocks) for tenant cluster Pods are changing. They now must adhere to the rules described in the [AWS documentation](https://docs.aws.amazon.com/vpc/latest/userguide/VPC_Subnets.html#add-cidr-block-restrictions).
+## aws-operator 8.2.2
```
MarcelMue, line 5: We usually have the `v` prefix to versions here. Linking to the `CHANGELOG.md` in the operator can come once it's finished.

release-notes/aws/v11.2.0.md
marians: Changed

release-notes/aws/v11.2.0.md
marians: Changed

aws.yaml
xh3b4sd: I would then like to also get a new version of `cluster-operator` in because there were at least the go modules changes. Releasing is easy, just creating the Github Release. Should be `2.1.6` then.

marians:
Note: target release version number is changing to v11.1.2

MarcelMue:
> Note: target release version number is changing to v11.1.2

Why? I mean AWS CNI seems like a pretty substantial change. Why not at least go with a minor?

marians:
> Why? I mean AWS CNI seems like a pretty substantial change. Why not at least go with a minor?

11.1.0 was introduced forAWS CNI and used as an internal test release (only for one customer). So we are consistent with having the minor upgrade from 11.0.x for every other customer.

PR #212, opened 2020-04-06
Add AWS release 11.1.3, deprecate 11.1.2

marians:
This publishes a fix in aws-operator.


README.md
```diff
@@ -6,7 +6,7 @@ This repository contains Giant Swarm releases and changelogs.
 
 ### AWS
 
- - [11.1.2](https://github.com/giantswarm/releases/blob/master/release-notes/aws/v11.1.2.md)
+ - [11.1.3](https://github.com/giantswarm/releases/blob/master/release-notes/aws/v11.1.3.md)
```
xh3b4sd, line 9: I think we should only add the new release.

release-notes/aws/v11.1.3.md
```diff
@@ -0,0 +1,50 @@
+# :zap: Giant Swarm Release v11.1.3 for AWS :zap:
+
+__Note:__ Upgrading to this release from any release prior 11.1.1 will cause a network downtime due to the network-related changes coming with the switch from Calico CNI to AWS CNI.
+
+## aws-operator [v8.2.2](https://github.com/giantswarm/aws-operator/releases/tag/v8.2.3)
```
xh3b4sd, line 5: This here should only contain the patched changes. 
```suggestion
## aws-operator [v8.2.3](https://github.com/giantswarm/aws-operator/releases/tag/v8.2.3)
```

README.md
marians: There is no point linking to the replaced release from the main README imo.

xh3b4sd: There is because we released something and now we release a patch. People use it already and we should reflect the history transparently.

marians: I'll add it back and may provide a cleanup PR in the future

PR #221, opened 2020-04-09
Create v11.2.3.md Release

T-Kukawka:


release-notes/azure/v11.2.3.md
```diff
@@ -0,0 +1,11 @@
+## ⚡️ Giant Swarm Release 11.2.3 for Azure ⚡️
+
+This release improves the way upgrades handle Azure VMSS Instances to make them more reliable in the Germany West Central location. This does not affect any existing functionality.
+
+To upgrade, please contact your Solution Engineer.
+
+Below, you can find more details on components that were changed with this release.
+
+### azure-operator v3.0.6
```
cokiengchiara, line 9: Just curious and should not be a blocker. I noticed other components link to the changelog. azure-operator does not. Any reason for that? See https://github.com/giantswarm/releases/blob/master/release-notes/azure/v11.2.1.md as an example.

T-Kukawka: good point, we will sort it when we release/test it with GWC

PR #222, opened 2020-04-09
Define and activate aws 9.0.2 release

sslavic:
Changes included:
- upgrade of nginx IC App from 1.6.5 to 1.6.8

### DIFF
--- a/release-notes/aws/v9.0.2.md
+++ b/release-notes/aws/v9.0.2.md
@@ -1,0 +1,18 @@
+## :zap: Giant Swarm Release 9.0.2 for AWS :zap:
+
+This release improves the reliability of NGINX Ingress Controller. Most important, termination configuration was adjusted so that active connections now get drained gracefully.
+
+Below, you can find more details on components that were changed with this release.
+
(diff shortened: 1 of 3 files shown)


release-notes/aws/v9.0.2.md
```diff
@@ -0,0 +1,18 @@
+## :zap: Giant Swarm Release 9.0.2 for AWS :zap:
+
```
cokiengchiara, line 2:
```suggestion

**If you are upgrading from 9.0.1, upgrading to this release will not roll your nodes. It will only update the apps.**

```

cokiengchiara: @sslavic , this applies here too right, because nginx version is defined here and not inside cluster operator?

sslavic: Well spotted, thank. Suggestion has been applied. Please take another look.

sslavic:
Tested on gauss:
- clean 9.0.2 installation
- upgrade from 9.0.1 to 9.0.2
- upgrade from 8.5.0 to 9.0.2

PR #225, opened 2020-04-14
deactivate AWS v11.0.1

xh3b4sd:
As discussed in https://gigantic.slack.com/archives/C7865GLSY/p1586723962074000 it looks like we can certainly deprecate `v11.0.1` since it does not make much sense to let users still create Tenant Clusters using the Calico CNI. There are stable releases using the AWS CNI.

### DIFF
--- a/aws.yaml
+++ b/aws.yaml
@@ -121,7 +121,7 @@ metadata:
     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
   name: v11.0.1
 spec:
-  state: active
+  state: deprecated
   date: 2020-02-05T12:00:00Z
   apps:
   - name: cert-exporter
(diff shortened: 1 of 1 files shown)


xh3b4sd:
I would also like to get approvals from @paurosello and @marians in case this is fine.

paurosello:
Sure, that sounds good but would like to manage the default PodCIDR in all installations first

xh3b4sd:
What is the ETA for that? How can I help with that?

xh3b4sd:
@paurosello can you answer my comment above? Where are we?

paurosello:
In Adidas and Shutterstock it has not been decided which IP CIDR will be used, and AWS CNI cluster creation is not possible there yet.

So we need to hold this for a while until this customers are able to use the AWS CNI release.

xh3b4sd:
Sure thing. Question was the ETA.

xh3b4sd:
@paurosello as discussed please check with the SEs and let's get this merged.

PR #232, opened 2020-04-21
KVM 9.0.1

stone-z:
pre-merge:
- [x] deprecate 9.0.0
- [x] remove -dev version from 9.0.1

### DIFF
--- a/kvm.yaml
+++ b/kvm.yaml
@@ -233,8 +233,8 @@ spec:
     componentVersion: 1.6.5
     version: 1.1.8
   - name: kube-state-metrics
-    componentVersion: 1.9.3
-    version: 1.0.5
+    componentVersion: 1.9.2
+    version: 1.0.4
   - name: metrics-server
     componentVersion: 0.3.3
     version: 1.0.0
--- a/kvm.yaml
+++ b/kvm.yaml
@@ -3,9 +3,60 @@ kind: Release
 metadata:
   annotations:
     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
-  name: v11.2.2
+  name: v11.2.3
 spec:
   state: active
+  date: 2020-04-23T12:00:00Z
+  apps:
+  - name: cert-exporter
+    version: 1.2.1
+  - name: chart-operator
+    version: 0.13.0
+  - name: coredns
+    componentVersion: 1.6.5
+    version: 1.1.8
+  - name: kube-state-metrics
+    componentVersion: 1.9.5
+    version: 1.0.5
+  - name: metrics-server
+    componentVersion: 0.3.3
+    version: 1.0.0
+  - name: net-exporter
+    version: 1.7.0
+  - name: nginx-ingress-controller
+    componentVersion: 0.30.0
+    version: 1.6.9
+  - name: node-exporter
+    componentVersion: 0.18.1
+    version: 1.2.0
+  components:
+  - name: app-operator
+    version: 1.0.0
+  - name: cert-operator
+    version: 0.1.0
+  - name: cluster-operator
+    version: 0.23.8
+  - name: flannel-operator
+    version: 0.2.0
+  - name: kvm-operator
+    version: 3.10.0
+  - name: kubernetes
+    version: 1.16.3
+  - name: containerlinux
+    version: 2191.5.0
+  - name: calico
+    version: 3.10.1
+  - name: etcd
+    version: 3.3.17
+---
+apiVersion: release.giantswarm.io/v1alpha1
+kind: Release
+metadata:
+  annotations:
+    giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
+  name: v11.2.2
+spec:
+  state: deprecated
   date: 2020-04-09T12:00:00Z
   apps:
   - name: cert-exporter
--- a/kvm.yaml
+++ b/kvm.yaml
@@ -284,8 +284,8 @@ spec:
     componentVersion: 1.6.5
     version: 1.1.8
   - name: kube-state-metrics
-    componentVersion: 1.9.5
-    version: 1.0.5
+    componentVersion: 1.8.0
+    version: 0.6.0
   - name: metrics-server
     componentVersion: 0.3.3
     version: 1.0.0
--- a/kvm.yaml
+++ b/kvm.yaml
@@ -273,7 +273,7 @@ metadata:
     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
   name: v9.0.0
 spec:
-  state: deprecated
+  state: active
   date: 2019-10-28T12:00:00Z
   apps: []
   components:
--- a/kvm.yaml
+++ b/kvm.yaml
@@ -324,7 +324,7 @@ metadata:
     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
   name: v9.0.0
 spec:
-  state: active
+  state: deprecated
   date: 2019-10-28T12:00:00Z
   apps: []
   components:
--- a/kvm.yaml
+++ b/kvm.yaml
@@ -256,7 +256,7 @@ spec:
   - name: flannel-operator
     version: 0.2.0
   - name: kvm-operator
-    version: 3.9.1
+    version: 3.9.1-dev
   - name: kubernetes
     version: 1.15.11
   - name: containerlinux
--- a/kvm.yaml
+++ b/kvm.yaml
@@ -264,7 +264,7 @@ spec:
   - name: calico
     version: 3.9.1
   - name: etcd
-    version: 3.4.7
+    version: 3.3.15
 ---
 apiVersion: release.giantswarm.io/v1alpha1
 kind: Release
--- a/kvm.yaml
+++ b/kvm.yaml
@@ -223,7 +223,7 @@ metadata:
   name: v9.0.1
 spec:
   state: active
-  date: 2020-04-20T12:00:00Z
+  date: 2020-04-24T12:00:00Z
   apps:
   - name: cert-exporter
     version: 1.2.2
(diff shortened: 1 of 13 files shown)


sslavic:
@tfussell @giantswarm/team-celestial same question with this one as with aws 9.0.3, would it make sense to ship kvm 9.0.1 with app changes only this week?

stone-z:
It seems like we're ready implementation-wise for 9.0.1 with flatcar and 1.15.11. The only remaining question is the cause of some inconsistencies in network behavior between our test installations and e.g. anubis. @MarcelMue @calvix have been looking into this

cokiengchiara:
> It seems like we're ready implementation-wise for 9.0.1 with flatcar and 1.15.11. The only remaining question is the cause of some inconsistencies in network behavior between our test installations and e.g. anubis. @MarcelMue @calvix have been looking into this

hi @giantswarm/team-ludacris , can Halo (cc @sslavic ) still contribute latest nginx ic and chart-operator to this release? (similar to https://github.com/giantswarm/releases/blob/master/release-notes/kvm/v11.2.3.md)

cokiengchiara:
@sslavic , if it is too disruptive for Ludacris, what if we just contribute changes to the next person doing a KVM release? Is any customer waiting for our changes here?

stone-z:
@cokiengchiara those versions are already in :)

release-notes/kvm/v9.0.1.md
```diff
@@ -0,0 +1,111 @@
...
+- Use upstream default of max-worker-connections of 16384.
+- Ignore NGINX IC Deployment replica count configuration when HorizontalPodAutoscaler is enabled.
+- Drop unnecessary Helm release revision annotation from NGINX IC Deployment.
+- Adjusted README for display in the web interface context.
+- HorizontalPodAutoscaler was tuned to use targetMemoryUtilizationPercentage of 80
+- Removed use of enable-dynamic-certificates CLI flag, it has been deprecated since ingress-nginx 0.26.0
+- Changed default error-log-level from error to notice.
+- Align graceful termination configuration with changes made in upstream ingress-nginx 0.26.0
+- Make controller.minReadySeconds configurable.
+- Change deployment to use release revision not time for Helm 3 support.
```
sslavic, line 101:
```suggestion
```
This was dropped, see above

> - Drop unnecessary Helm release revision annotation from NGINX IC Deployment.

release-notes/kvm/v9.0.1.md
```diff
@@ -0,0 +1,110 @@
...
+
+- Updated for Kubernetes 1.16 compatibility.
+
+### nginx-ingress-controller v0.30.0 ([Giant Swarm app v1.6.9](https://github.com/giantswarm/nginx-ingress-controller-app/blob/master/CHANGELOG.md#v169-2020-04-22))
+
+- Restrict PodSecurityPolicy volumes to only those required (removes wildcard).
+- Tune net.ipv4.ip_local_port_range to 1024 65535 as a safe sysctl.
+- Tune net.core.somaxconn to 32768 via an initContainer with privilege escalation.
+- Use 4 worker processes by default.
+- Use upstream default of max-worker-connections of 16384.
```
sslavic, line 92:
```suggestion
```
This was prior state, got changed in between and now restored again with latest nginx app release, so for customer upgrading from kvm 9.0.0 there was no actual change.

kvm.yaml
```diff
@@ -217,6 +217,57 @@ spec:
...
+  - name: cert-exporter
+    version: 1.2.2
+  - name: chart-operator
+    version: 0.13.0
+  - name: coredns
+    componentVersion: 1.6.5
+    version: 1.1.8
+  - name: kube-state-metrics
+    componentVersion: 1.9.2
+    version: 1.0.4
```
sslavic, line 237:
```suggestion
    componentVersion: 1.9.5
    version: 1.0.5
```
Are you sure KSM 1.0.5/1.9.5 is not compatible with k8s 1.15? Seems to work well for equivalent aws release https://github.com/giantswarm/releases/blob/master/aws.yaml#L711-L761

If suggestion is accepted, release notes for 1.0.5 have to be added too.

stone-z: According to their compatibility matrix, 1.9.5 is k8s 1.16. I was running it accidentally with this release for a while without noticeable issues, but unless there's a big benefit for us I'd keep it at the supported version. This is already a lot of change

sslavic: There is big benefit - 1.9.5 works while 1.9.2 is crashing and paging https://github.com/giantswarm/giantswarm/issues/9506

It's compatible with 1.16 and it works well with 1.15 too.

stone-z: Alright, if we know it works I'll go ahead and add it back

MarcelMue: For us when testing it did not work on kvm 1.15 - we can try again but it was full of error logs when we tried.

stone-z: Yes @sslavic looks like there is one of these per second in the ksm logs:
```
E0424 07:47:30.318475       1 reflector.go:156] pkg/mod/k8s.io/client-go@v0.0.0-20191109102209-3c0d1af94be5/tools/cache/reflector.go:108: Failed to list *v1.ValidatingWebhookConfiguration: the server could not find the requested resource
```

stone-z: It's definitely a change on the upstream side 
https://github.com/kubernetes/kube-state-metrics/issues/1093

Honestly, to me this doesn't seem as bad as ksm crashlooping, but would this cause other issues? Are these logs monitored? Would this fill up a volume?

sslavic: Thanks guys for heads up, well spotted, testing like https://gigantic.slack.com/archives/C90BZMUJ1/p1587651052118200?thread_ts=1587390760.075200&cid=C90BZMUJ1 cannot spot that kind of issue, would be great to have this testing automated.

Found related issue / support request and suggestion to use 1.8.* for 1.15:
https://github.com/kubernetes/kube-state-metrics/issues/1093#issuecomment-596837244
which is aligned with the [compatibility matrix](https://github.com/kubernetes/kube-state-metrics#compatibility-matrix).

This probably affects all supported 1.15 based releases. We practically need separate ksm 1.15 compatibility branch if none of existing releases works. In that branch (if turned out to be need) we should still try to make all resources (other than binary) compatible with k8s 1.16 so that we have no problem with stuck helm releases during upgrade to k8s 1.16 based release.

Noting once more, 1.9.2 is not an option as well since it's crashing and paging https://github.com/giantswarm/giantswarm/issues/9506 and by the matrix 1.9.* binaries are not compatible with k8s 1.15

MarcelMue: So can we just keep it as the same version we used for `9.0.0`?
https://github.com/giantswarm/releases/blob/master/release-notes/kvm/v9.0.0.md#kube-state-metrics-v180-gs-v060

stone-z: I deployed a new release version with the old KSM on gorgoth. KSM logs have no errors, only one warning:
`W0424 09:05:19.624035       1 client_config.go:541] Neither --kubeconfig nor --master was specified.  Using the inClusterConfig.  This might not work.`

stone-z: I've updated this release to include 1.8.0 / GS 0.6.0 kube-state-metrics

sslavic: It solves for having binary compatible with k8s 1.15, I'm concerned that it could cause upgrade issues later - resources in the ksm 0.6.0 chart may be not compatible with k8s 1.16, and helm release could end up being stuck, only recovery being to delete and re-install it post upgrade.
Since ksm is more relevant to us than it us to customer, even that concern is not an issue/blocker (workaround with recreate is acceptable mitigation), as long as issue can be detected at least. Prerequisite for that is to have ksm specific AppFailed alert defined and page or at least notify. See more related alerts https://github.com/giantswarm/g8s-prometheus/blob/master/helm/g8s-prometheus/prometheus-rules/app.rules.yml That can be done as followup. Please consider adding a todo for this in team ludacris board.

release-notes/kvm/v9.0.1.md
```diff
@@ -0,0 +1,103 @@
...
+To continue receiving security updates and to minimize the effort needed to migrate in the future, we recommend upgrading to this release.
+
+This release includes multiple improvements to the NGINX Ingress Controller app:
+1. It upgrades to upstream ingress-nginx `v0.30.0`.
+2. Optional metrics service for prometheus-operator support was added. This allows prometheus-operator to scrape metrics to monitor the app.
+3. NGINX IC SSL support for old browsers and clients (e.g. Safari 9) is dropped. This is due to default SSL ciphers no longer including AES-CBC based ciphers since they are considered weak. At your own risk, weak ciphers can still be enabled on demand independently for each cluster.
+
+Below, you can find more details on components that were changed with this release.
+
+### Kubernetes [v1.15.11](https://github.com/kubernetes/kubernetes/blob/master/CHANGELOG/CHANGELOG-1.15.md#changelog-since-v11510)
```
MarcelMue, line 19: Would it also make sense to include a changelog for `containerlinux`?
https://github.com/giantswarm/releases/blob/master/release-notes/kvm/v9.0.0.md#coreos-container-linux-v219150

stone-z: Yes good idea - included

stone-z:
Slack notification markdown:
```
:zap: Giant Swarm Release 9.0.1 for KVM is now active for you! :zap:

This release includes Kubernetes v1.15.11 as well as some reliability and user experience improvements.

In addition, this release replaces CoreOS with <https://www.flatcar-linux.org/|Flatcar Container Linux>. CoreOS has gone <https://coreos.com/os/eol/|end-of-life> and is being rapidly phased out. Flatcar is a compatible fork of CoreOS which receives ongoing support. To continue receiving security updates and to minimize the effort needed to migrate in the future, we recommend upgrading to this release.

This release includes a breaking change to the NGINX Ingress Controller which drops support for older browsers using weak SSL cipher suites. Contact your Solution Engineer for further details.

This is also the first KVM release which is internally represented by our <https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/|Release CRD>. This is done in preparation of opening up the control plane to you.

Read the full release notes here: https://github.com/giantswarm/releases/blob/master/release-notes/kvm/v9.0.1.md

```

PR #234, opened 2020-04-22
Define and activate AWS 9.2.4 release

sslavic:
Towards https://github.com/giantswarm/giantswarm/issues/10227

### DIFF
--- a/aws.yaml
+++ b/aws.yaml
@@ -363,9 +363,61 @@ kind: Release
 metadata:
   annotations:
     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
-  name: v9.2.3
+  name: v9.2.4
 spec:
   state: active
+  date: 2020-04-22T12:00:00Z
+  apps:
+  - name: cert-exporter
+    version: 1.2.1
+  - name: chart-operator
+    version: 0.12.1
+  - name: cluster-autoscaler
+    componentVersion: 1.16.2
+    version: 1.1.4
+  - name: coredns
+    componentVersion: 1.6.5
+    version: 1.1.8
+  - name: kube-state-metrics
+    componentVersion: 1.9.5
+    version: 1.0.5
+  - name: metrics-server
+    componentVersion: 0.3.3
+    version: 1.0.0
+  - name: net-exporter
+    version: 1.7.0
+  - name: nginx-ingress-controller
+    componentVersion: 0.30.0
+    version: 1.6.9
+  - name: node-exporter
+    componentVersion: 0.18.1
+    version: 1.2.0
+  components:
+  - name: app-operator
+    version: 1.0.0
+  - name: aws-operator
+    version: 5.6.0
+  - name: cert-operator
+    version: 0.1.0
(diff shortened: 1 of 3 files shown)


release-notes/aws/v9.2.4.md
```diff
@@ -0,0 +1,18 @@
+## :zap: Giant Swarm Release 9.2.4 for AWS :zap:
+
+**If you are upgrading from 9.2.3, upgrading to this release will not roll your nodes. It will only update the apps.**
+
+This release improves the reliability of NGINX Ingress Controller. Most importantly, kernel and app settings have been tuned to increase out-of-the-box performance, and app Helm chart has been adjusted to improve service availability (minimize disruption) during configuration changes rollout.
```
cokiengchiara, line 5:
```suggestion
This release improves the reliability of NGINX Ingress Controller. Most importantly, kernel and app settings have been tuned to increase out-of-the-box performance. The app's Helm chart was also adjusted to improve its availability when rolling out configuration changes.
```

aws.yaml
```diff
@@ -363,9 +363,61 @@ kind: Release
...
-  name: v9.2.3
+  name: v9.2.4
 spec:
   state: active
+  date: 2020-04-22T12:00:00Z
+  apps:
+  - name: cert-exporter
+    version: 1.2.1
+  - name: chart-operator
+    version: 0.12.1
```
cokiengchiara, line 374:
```suggestion
    version: 0.13.0
```

@sslavic @rossf7 should we also include latest chart-operator here? https://github.com/giantswarm/chart-operator/blob/master/CHANGELOG.md#v0130-2020-04-21

cokiengchiara:
> Also, does it address Adidas concerns? (I am not sure as to the specifics of those concerns)

For example, @yasn77 mentioned, "all replicas die pretty much the same time" do we want to address more directly here to alleviate their concerns? Just an idea, not a prescription.

sslavic:
@cokiengchiara thanks for review, applied your suggestions, PTAL

And yes I believe it does address Adidas concerns, at least from our joint learnings so far; there are more improvements we will do (besides the hard parts) like using Local external traffic policy by default for nginx NodePort and LoadBalancer Services (this is also relevant for dvag). Currently included changes are most critical and we'd all be more confident with next Saturday sales event to be run on this release. @yasn77 WDYT?

cokiengchiara:
> And yes I believe it does address Adidas concerns

Then that part LGTM. Waiting on your feedback re: chart-operator https://github.com/giantswarm/releases/pull/234/files#r412821436 

Then will post in SIG product for further feedback.

aws.yaml
cokiengchiara: No time to include it. Let's do it in 9.2.5.

sslavic: Sorry, too late for this one //cc @rossf7

sslavic:
This was tested on ginger and giraffe:
- create clean 9.2.4 cluster
- upgrade from 9.2.3 to 9.2.4
- upgrade from 9.0.2 to 9.2.4

@piontec also load tested nginx 1.6.9 on ginger

PR #231, opened 2020-04-21
Add AWS release v11.2.0

marians:
For https://github.com/giantswarm/giantswarm/issues/9888

PR #239, opened 2020-04-23
Fix kube-state-metrics versions in AWS releases

marians:
Follow-up to https://github.com/giantswarm/releases/pull/231

This fixes an error in the kube-state-metrics versions of v11.1.4 and 11.2.0.


marians:
Thanks for merging!

PR #263, opened 2020-05-04
drop unused releases

xh3b4sd:
Towards https://github.com/giantswarm/giantswarm/issues/10420. Please check selectively if this makes sense so that we verify and have consensus this is fine.


aws.yaml
```diff
@@ -1040,130 +504,6 @@ spec:
...
-    version: 3.7.2
-  - name: etcd
-    version: 3.3.13
----
-apiVersion: release.giantswarm.io/v1alpha1
-kind: Release
-metadata:
-  annotations:
-    giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
-  name: v8.2.1
```
marians, line 1140: Not sure about removing v8.2.1, as we still have cluster using v8.2.0.

marians:
To me it is super hard to read from the diff view what is actually going on in this PR. :( Example:

image

@xh3b4sd Can you add a list of what you are doing to which release, please?

aws.yaml
xh3b4sd: But the release is deprecated. Why should we bother with that?

xh3b4sd:
With you. I did all this by hand and put the list of unused release versions in the issues linked above. You do not need to confirm a 100% accuracy. You could try to verify by best effort. In worst case a used release is removed and we add it back. No big issue for me right now. I would go with this and iterate if necessary.

PR #267, opened 2020-05-05
Define and activate release 11.3.0 for AWS

tfussell:


aws.yaml
```diff
@@ -1,5 +1,63 @@
...
+  - name: cluster-operator
+    version: 2.1.10
+  - name: kubernetes
+    version: 1.16.9
+  - name: containerlinux
+    version: 2345.3.1
+  - name: calico
+    version: 3.10.1
+  - name: etcd
+    version: 3.3.17
```
marians, line 57: Please also add component `name: aws-cni` with `version: 1.6.0`. We just added that in retrospect for informational purposes to the releases making use of it.

tfussell: Done!

aws.yaml
```diff
@@ -1,5 +1,65 @@
...
+  - name: node-exporter
+    componentVersion: 0.18.1
+    version: 1.2.0
+  components:
+  - name: app-operator
+    version: 1.0.0
+  - name: aws-cni
+    version: 1.6.0
+  - name: aws-operator
+    version: 8.4.1-dev
```
xh3b4sd, line 47:
```suggestion
    version: 8.5.0
```

PR #273, opened 2020-05-06
Inform about the consequence of app cr migration

oponder:
This informs people that their `nginx-ingress-controller-user-values` have moved to the control plane.

Towards: https://github.com/giantswarm/giantswarm/issues/10211

fyi @giantswarm/se this is something to keep in mind, I guess you all know it by now due to some of the threads and incidents, but I am currently trying to get this down on paper somewhere for people that either forget it or never heard about it.

I'm figuring out also where it happens for KVM and Azure. Took me a while to wrap my head around the releases and how it all worked, but I think I have a grasp on it now.

### DIFF
--- a/release-notes/aws/v9.0.1.md
+++ b/release-notes/aws/v9.0.1.md
@@ -2,7 +2,19 @@
 
 This release includes Kubernetes v1.15.11 as well as some reliability and user experience improvements.
 We highly recommend you to upgrade to this release if you want to continue running on Kubernetes 1.15 for now.
+
 This is also the first release which is internally represented by our [Release CRD](https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/). This is done in preparation of opening up the control plane to you.
+
+---
+
+**Note** If you are upgrading from 8.5.0 or 9.0.0: The `nginx-ingress-controller-user-values` configmap has been migrated from the `kube-system` namespace of your Tenant Cluster to a namespace on the Control Plane.
+
+If you had a cluster with id `e05c8`, then you'll find your `nginx-ingress-controller-user-values`
+configmap now in a namespace called `e05c8` on the Control Plane.
+
+If you do not yet have access to the Control Plane API you will not be able to manually configure your ingress anymore. Please contact your Solution Engineer for more information.
+
+---
 
 This release includes multiple improvements to the NGINX Ingress Controller app:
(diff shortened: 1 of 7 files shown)


sslavic:
Auto migration of user values ConfigMap from TC to CP is not nginx specific. Underlying delivery pipeline (cluster-operator) manages it. All apps, pre-installed or optional are affected. Recently (@omegas27 can verify) after cluster upgrade coredns was not working as expected, customer (dvag) was surprised not to find coredns user values ConfigMap on TC anymore, they were overriding it, override was not working, they couldn't debug themselves, required support.

Lets not associate any more negative feelings to nginx service.

oponder:
Just realised that `coredns` and `cluster-autoscaler` would also need the same type of notice. Possibly other apps.

oponder:
@sslavic I will edit the message to not be nginx specific. will do so tomorrow morning while I gather a bit more info and possibly make this PR also correct for KVM and Azure

sslavic:
Maybe once https://github.com/giantswarm/giantswarm/issues/10211 is fixed, user values CM migration is not as breaking change as it was, at least maybe customers could continue to manage user values on TC, they would disappear from TC immediately, but would get applied. Customer deployment pipeline for user values CM might still have to be adjusted not to panic or redeploy repeatedly when user values CM gets migrated to CP.

oponder:
Yeah giantswarm/giantswarm#10211 getting merged will help the outages caused by customers still trying to apply configuration in the old way.

But what I want to address is the awareness that this upgrade needs a change customer workflows.

sslavic:
True, unwanted breaking change ideally should be avoided, but since it happened at least needs to be communicated. Somehow we missed to add changelog entry to cluster-operator that included the migration logic, then also in releases which use it.

Customers have to adapt their deployment pipeline of user values ConfigMap and Apps; to make that possible they need to have k8s CP opened up for them and for that they need to reach out to SEs, since AFAIK it's not self-service.

oponder:
So the notice makes sense right? (aside from the fact that it is NOT nginx specific, I'll fix that. was never my intention to associate any negativity there. I want to bring clarity)

rossf7:
Including in the release notes makes sense to me. The migration process was designed so it is automated but the change in location of the configuration was not communicated clearly to SEs or customers.

There was also the bug in the migration logic that triggered the PM and had a big impact. The fix is prepared and thanks Oliver for helping test it and for improving the communication.

PR #276, opened 2020-05-07
Add aws-cni component info

marians:
Towards https://github.com/giantswarm/giantswarm/issues/10700

This adds info on a component that was part of the releases, but not contained as such. To show more complete information to end users, we add it here in retrospect.

FYI: `aws-cni` is a name we made up and put up here for discussion. An alternative would be the default image name when buildig the original image from [source](https://github.com/aws/amazon-vpc-cni-k8s): `amazon-k8s-cni`.


xh3b4sd:
We should have a story for Firecracker to wire it accordingly. This here is now only cosmetic but should then rather be used for the real configuration in the backend, just like we do with the other versions.

PR #277, opened 2020-05-07
Separate versions

tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/8655


aws/kustomization.yaml
```diff
@@ -0,0 +1,26 @@
+commonAnnotations:
+  giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
+
+resources:
```
MarcelMue, line 4: Can we somehow generate this? We have to duplicate the info otherwise and this is very easy to go wrong.

README.md
```diff
@@ -13,92 +13,92 @@ Swarm Control Planes.
 
 - v11
   - v11.2
-    - [v11.2.1](https://github.com/giantswarm/releases/blob/master/release-notes/aws/v11.2.1.md)
-    - [v11.2.0](https://github.com/giantswarm/releases/blob/master/release-notes/aws/v11.2.0.md)
+    - [v11.2.1](https://github.com/giantswarm/releases/blob/master/aws/v11.2.1/release-notes.md)
```
MarcelMue, line 16: This is an issue - this will break all the links in `slack` we have pushed out to customers right now :(

Maybe symlink or something? Even though I doubt it works as we direct link to the `.md`.

.circleci/config.yml
```diff
@@ -5,6 +5,23 @@ orbs:
...
 jobs:
+  build:
+    docker:
+      - image: quay.io/giantswarm/architect
+    steps:
+      - checkout
+      - run:
+          name: Install kustomize
+          command: |
+            CGO_ENABLED=0 go install sigs.k8s.io/kustomize/kustomize/v3
```
MarcelMue, line 16: Is there no `kubectl` in here? `kustomize` is included in kubectl version 1.16+ IIRC

tfussell: I didn't see it. Adding kubectl/kustomize to architect would be an option though. I'll confirm.

README.md
tfussell: I was worried about that. Using links directly to GitHub ties us to this directory structure. I could keep the notes where they are. Or we could have a migration period. Or we could host the notes with a real website where we could do redirects though that wouldn't help this particular case.

aws/kustomization.yaml
tfussell: Agreed. As I was moving things around I realized we could have a generator, maybe `devctl create release` which initializes the directory and updates the root list. It could base off of another release.

MarcelMue: Could be one idea to improve this :+1:

.circleci/config.yml
MarcelMue: Not a massive deal either way.

ubergesundheit: Keep in mind the kustomize version in kubectl is older than the standalone version

README.md
tfussell: After some discussion, it sounds like broken links in Slack aren't a high priority.

aws/kustomization.yaml
tfussell: For now I will create a checklist so we don't miss one of these manual steps.

.circleci/config.yml
tfussell: I don't see `kubectl` in `architect` and I don't want to bring it in just for this job so I'll stick with this approach if that's alright.

aws/kustomization.yaml
tfussell: I actually turned the checks into unit tests so it will be impossible to make these mistakes.

tfussell:
I will manually confirm that no releases are lost/modified in this PR before merging.

oponder:
Adding a todo I identified:

- [ ] Update docs here to mention that releases are in individual files now? https://intranet.giantswarm.io/docs/dev-and-releng/releases/

xh3b4sd:
I wanted to give a little feedback on further improvements. It is quite difficult to tell what changes in a release. So it would be cool to have some means to render or show a diff to the previous version. Then one can easily tell and review what the successor release is actually about. Maybe we could render a diff file and fail CI if it is not available.

tfussell:
> Adding a todo I identified:
> 
> * [ ]  Update docs here to mention that releases are in individual files now? https://intranet.giantswarm.io/docs/dev-and-releng/releases/

Thanks, I haven't explored the intranet much yet. It's not easy to know all of the docs that might need to be updated for a change.

tfussell:
> I wanted to give a little feedback on further improvements. It is quite difficult to tell what changes in a release. So it would be cool to have some means to render or show a diff to the previous version. Then one can easily tell and review what the successor release is actually about. Maybe we could render a diff file and fail CI if it is not available.

This would be cool. I can think of a few different approaches. One would be an "event sourcing" model where a release can define a base and only changed components (effectively a diff). You could then follow the bases backwards to get the full list of component and app versions. As a bonus, you would have a defined chain of allowed/tested upgrades. In the following example, if 11.2.2 is released after 11.3.0, it would still be clear that 11.3.0 is based on 11.2.1 and and an upgrade from 11.2.2 to 11.3.0 might not be tested.

```
apiVersion: release.giantswarm.io/v1alpha1
kind: Release
metadata:
  annotations:
    giantswarm.io/base-release: v11.2.1
  name: v11.3.0
spec:
  state: active
  date: 2020-05-11T15:00:00Z
  components:
  - name: aws-operator
    version: 8.5.0
  - name: kubernetes
    version: 1.16.9
  - name: containerlinux
    version: 2345.3.1
```
A simpler option would be to provide some scripts that compare two releases (also using some specified base release) and create a skeleton of release notes. For example, if we create a new release that changes only `cluster-operator` version, the release notes should look like:
```
## :zap: Giant Swarm Release <version> for <provider> :zap:

<Description here>

### cluster-operator [<cluster operator version>](https://github.com/giantswarm/cluster-operator/releases/tag/<cluster operator version>)
- <changes>
```
This is something we could check with CI so all changes are clear to reviewers and inadvertent updates don't sneak in.

xh3b4sd:
> an "event sourcing" model where a release can define a base

I would like to stick with the full yaml because you can easily apply it. I think the order is imperative by the actual release version, which is semver. We just need something that compares two releases and finds the diff so it can be rendered in some format.

tfussell:
I finally found the focus to check this PR. Here's a little script I wrote which splits `aws.yaml`/`azure.yaml`/`kvm.yaml` from `master` into separate release CRs using [kubernetes-split-yaml](https://github.com/mogensen/kubernetes-split-yaml), moves release CRs in this branch into the same directory structure, and compares the directories with `diff` to make sure there are no mistakes. Here is the script if anyone wants to check:
```bash
for i in aws azure kvm; do
  echo $i
  wget -q https://raw.githubusercontent.com/giantswarm/releases/master/$i.yaml -O $i.yaml
  $GOPATH/bin/kubernetes-split-yaml $i.yaml 2> /dev/null
  rm -rf $i-master
  mv generated $i-master
  rm -rf $i-branch
  mkdir $i-branch
  for j in $(ls $i); do
    if [[ "$j" == "kustomization.yaml" ]]; then
      continue
    fi
    cp $i/$j/release.yaml $i-branch/$j-release.yaml
  done
  diff $i-master $i-branch
done
```
and here is the output (annotations are added with `kustomize` so it's not really lost):
```diff
aws
Only in aws-branch: v10.1.0-release.yaml
Only in aws-branch: v10.1.1-release.yaml
Only in aws-branch: v10.1.2-release.yaml
Only in aws-branch: v11.0.0-release.yaml
diff aws-master/v11.0.1-release.yaml aws-branch/v11.0.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
Only in aws-branch: v11.1.2-release.yaml
Only in aws-branch: v11.1.3-release.yaml
diff aws-master/v11.1.4-release.yaml aws-branch/v11.1.4-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v11.2.0-release.yaml aws-branch/v11.2.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v11.2.1-release.yaml aws-branch/v11.2.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v11.3.0-release.yaml aws-branch/v11.3.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v8.2.0-release.yaml aws-branch/v8.2.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v8.5.0-release.yaml aws-branch/v8.5.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v9.0.0-release.yaml aws-branch/v9.0.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
Only in aws-branch: v9.0.1-release.yaml
Only in aws-branch: v9.0.2-release.yaml
diff aws-master/v9.0.3-release.yaml aws-branch/v9.0.3-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v9.0.4-release.yaml aws-branch/v9.0.4-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v9.0.5-release.yaml aws-branch/v9.0.5-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
Only in aws-branch: v9.1.0-release.yaml
Only in aws-branch: v9.2.0-release.yaml
Only in aws-branch: v9.2.1-release.yaml
diff aws-master/v9.2.2-release.yaml aws-branch/v9.2.2-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
Only in aws-branch: v9.2.3-release.yaml
diff aws-master/v9.2.4-release.yaml aws-branch/v9.2.4-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v9.2.5-release.yaml aws-branch/v9.2.5-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v9.2.6-release.yaml aws-branch/v9.2.6-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v9.3.0-release.yaml aws-branch/v9.3.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v9.3.1-release.yaml aws-branch/v9.3.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff aws-master/v9.3.2-release.yaml aws-branch/v9.3.2-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
azure
diff azure-master/v11.0.0-release.yaml azure-branch/v11.0.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v11.1.0-release.yaml azure-branch/v11.1.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v11.2.0-release.yaml azure-branch/v11.2.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v11.2.1-release.yaml azure-branch/v11.2.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v11.2.2-release.yaml azure-branch/v11.2.2-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v11.2.3-release.yaml azure-branch/v11.2.3-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v11.2.4-release.yaml azure-branch/v11.2.4-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v11.2.5-release.yaml azure-branch/v11.2.5-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v11.2.6-release.yaml azure-branch/v11.2.6-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v11.3.0-release.yaml azure-branch/v11.3.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v11.3.1-release.yaml azure-branch/v11.3.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v8.0.0-release.yaml azure-branch/v8.0.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v8.2.0-release.yaml azure-branch/v8.2.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v8.2.1-release.yaml azure-branch/v8.2.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v8.3.0-release.yaml azure-branch/v8.3.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v8.4.0-release.yaml azure-branch/v8.4.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v8.4.1-release.yaml azure-branch/v8.4.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff azure-master/v9.0.0-release.yaml azure-branch/v9.0.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
kvm
diff kvm-master/v11.0.0-release.yaml kvm-branch/v11.0.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v11.1.0-release.yaml kvm-branch/v11.1.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v11.2.0-release.yaml kvm-branch/v11.2.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v11.2.1-release.yaml kvm-branch/v11.2.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v11.2.2-release.yaml kvm-branch/v11.2.2-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v11.2.3-release.yaml kvm-branch/v11.2.3-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v11.3.0-release.yaml kvm-branch/v11.3.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v11.3.1-release.yaml kvm-branch/v11.3.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v8.0.0-release.yaml kvm-branch/v8.0.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v8.1.0-release.yaml kvm-branch/v8.1.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v8.2.0-release.yaml kvm-branch/v8.2.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v8.2.1-release.yaml kvm-branch/v8.2.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v8.3.0-release.yaml kvm-branch/v8.3.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v8.4.0-release.yaml kvm-branch/v8.4.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v9.0.0-release.yaml kvm-branch/v9.0.0-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v9.0.1-release.yaml kvm-branch/v9.0.1-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v9.0.2-release.yaml kvm-branch/v9.0.2-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
diff kvm-master/v9.0.3-release.yaml kvm-branch/v9.0.3-release.yaml
4,5d3
<   annotations:
<     giantswarm.io/docs: https://docs.giantswarm.io/reference/cp-k8s-api/releases.release.giantswarm.io/
```

tfussell:
Here is the helm chart for the most recent commit https://github.com/giantswarm/releases-test-catalog/commit/507a79dad55181456ebfc0fa59549cde8a57a480 It's not very easy to compare with the existing `aws.yaml` due to reordering though.

xh3b4sd:
Does there anything change in the process or organization with this? Do people know how to work with releases when the structure changes? Is this written down/communicated?

tfussell:
> Does there anything change in the process or organization with this? Do people know how to work with releases when the structure changes? Is this written down/communicated?

I won't merge this until I have some docs written up. I will also announce it and provide support for upcoming releases to ease the transition.

tfussell:
Documented in https://github.com/giantswarm/giantswarm/pull/11393 I don't expect any issues from merging this, but I will wait until tomorrow morning when others are online to be safe.

marians:
Just want to express my ❤️  for this PR!

marians:
Too late for this PR, but I had the though: did you consider putting the release notes into a `README.md` file instead of `release-notes.md`? That would have two benefits:

- When navigating folders via GitHub, the release notes would be displayed as soon as you're in the folder for a release.
- Release notes URLs, which are now like `https://github.com/giantswarm/releases/blob/master/{provider}/{version}/release-notes.md` could be a bit shorter, like `https://github.com/giantswarm/releases/blob/master/{provider}/{version}`

tfussell:
> Too late for this PR, but I had the thought: did you consider putting the release notes into a `README.md` file instead of `release-notes.md`?

I like it. I'll make a note in our team board and probably knock it out next week.

PR #293, opened 2020-05-21
Vaclav 12.0.1 hama

calvix:


aws.yaml
```diff
@@ -1,5 +1,65 @@
...
+  - name: cert-exporter
+    version: 1.2.1
+  - name: cert-manager
+    componentVersion: 0.9.0
+    version: 1.0.7
+  - name: chart-operator
+    version: 0.12.4
+  - name: cluster-autoscaler
+    componentVersion: 1.16.2
+    version: 1.1.4
```
marians, line 20:
```suggestion
    componentVersion: 1.16.5
    version: 1.16.0
```

If you'd like to test the latest cluster-autoscaler version along the way...

PR #296, opened 2020-05-26
Remove commented out chart-operator

rossf7:
@paurosello Not urgent but just noticed this when I was testing cluster-operator.


marians:
I think in master this has been removed in the meantime.

rossf7:
Thanks @marians

PR #313, opened 2020-06-08
Archive old releases

tfussell:
While separating releases into directories, I brought in some previously disabled releases which we would like to keep for historical purposes but they should not be available on installations. I checked the versions of all TCs on all installations and found that these releases are not currently being used. A second set of eyes to check these would be great.


tfussell:
@giantswarm/team-ludacris @giantswarm/team-celestial @giantswarm/team-firecracker Please take a look at your provider releases to see if anything looks incorrect if you have a moment.

azure/kustomization.yaml
```diff
@@ -2,21 +2,8 @@ commonAnnotations:
...
   - v9.0.0
-  - v11.0.0
-  - v11.1.0
-  - v11.2.0
-  - v11.2.1
-  - v11.2.2
-  - v11.2.3
-  - v11.2.4
-  - v11.2.5
   - v11.2.6
```
nprokopic, line 7: You can also archive `11.2.6` for Azure.

PR #314, opened 2020-06-09
Add note for changed user values configmap

cokiengchiara:
These two notes should persist across all 9.0.x releases to prevent customer automation from causing issues. @tomahawk28 this is correct and I'm not saying some bullshit right? Haha


tomahawk28:
> These two notes should persist across all 9.0.x releases to prevent customer automation from causing issues. 

You're right, Since 9.0.6 is the first version from 9.x.x version, we should have this note here. 
Thanks for the notice!

cokiengchiara:
> It's a little surprising for me to have a message for our SEs in release notes since these are usually targeted towards customers, but I don't mind if you think it's useful. The content is good.

Thanks. Same here. Right now, we have no reliable place for SEs to look for these kinds of warnings. They are relevant for customers too. So I think it works for now. I checked in chapter-se channel too and @pipo02mix agreed it might help :)

PR #336, opened 2020-07-06
Allow suffixes in release names

yulianedyalkova:
This is part of the [release-operator implementation](https://github.com/giantswarm/giantswarm/issues/11716) and it will allow people to create test releases by simply adding a suffix to the name. So I will be able to call my test release v13.3.0-yulia instead of coming up with some random numbers.


releases_test.go
```diff
@@ -141,30 +141,33 @@ func Test_Releases(t *testing.T) {
...
-			err = v1beta1.Convert_v1beta1_CustomResourceValidation_To_apiextensions_CustomResourceValidation(crd.Spec.Validation, &v, nil)
-			if err != nil {
-				t.Fatal(microerror.Mask(err))
-			}
 
-			validator, _, err := validation.NewSchemaValidator(&v)
-			if err != nil {
-				t.Fatal(microerror.Mask(err))
-			}
+			for _, crdVersion := range crd.Spec.Versions {
```
MarcelMue, line 145: Should this iterate through or just take the highest version in there?

yulianedyalkova: Tbh I'm not sure, I just adapted the tests in the same way that we did it in [apiextensions](https://github.com/giantswarm/apiextensions/commit/1c32c5de067017dc8fe52dc718543b8b5e5e7179). 

Note that those are crd versions (so this should be just v1 in our case I think?) and not release versions.

tfussell: In this case, these are versions of the Releases API defined by the CRD rather than the version of the CRD itself (v1/v1beta1). It's possible to have different OpenAPIV3 schemata for different versions which is why I added this loop. Currently we only have one version of Releases so it's not that important either way.

PR #340, opened 2020-07-07
Sort releases and kustomization.yaml resources

tfussell:
To enable automated release creation, we need a consistent format for releases. I wrote a quick script which reads all releases and writes them to the same file with sorted fields and sorted components. I also sorted aws/azure/kvm kustomization.yaml for the same reason. There should be no change to the content of the releases.

### DIFF
--- a/aws/archived/v10.1.1/release.yaml
+++ b/aws/archived/v10.1.1/release.yaml
@@ -1,29 +1,30 @@
 apiVersion: release.giantswarm.io/v1alpha1
 kind: Release
 metadata:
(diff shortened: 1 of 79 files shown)


tfussell:
Tested on `geckon`. I didn't see any issues with the releases. Aside from reordering of components/apps, everything looks the same.

MarcelMue:
> Tested on `geckon`. I didn't see any issues with the releases. Aside from reordering of components/apps, everything looks the same.

Hit the button.

PR #325, opened 2020-06-23
Release AWS 11.4.1

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/11946

PR #346, opened 2020-07-12
Fix external-dns component version in aws 11.4.1

sslavic:
AWS 11.4.1 includes external-dns App v1.2.2 which has upgrade to upstream external-dns 0.7.2 see https://github.com/giantswarm/releases/pull/325

Metadata about external-dns component version in the platform release has not been updated. This PR fixes the issue.


sslavic:
Going with a ping.

PR #337, opened 2020-07-06
Define and activate AWS 9.3.4 release

sslavic:
Towards giantswarm/giantswarm#11530

PR #350, opened 2020-07-13
Fix cluster-autoscaler component version metdata in aws 9.3.4

sslavic:
Missed to update this in https://github.com/giantswarm/releases/pull/337


sslavic:
Going with a ping

PR #360, opened 2020-07-20
Add TWIR Jul 24

cokiengchiara:
Open this week's releases.


announcements/2020-Week-30/aws.md
```diff
@@ -0,0 +1,13 @@
+# :zap: This Week in AWS Releases (July 24, 2020):zap:
+
+[AWS v9.3.5](https://github.com/giantswarm/releases/tree/master/aws/v9.3.5) updates NGINX Ingress Controller to the latest release and fixes a regression introduced in the previous upstream version.
```
sslavic, line 3: We included also upgrade to k8s 1.16.12 (see https://github.com/giantswarm/releases/tree/master/aws/v9.3.5#kubernetes-11612)

Should it be covered here? //cc @MartaD

MartaD: Yes, thank you, it’s there now.

sslavic:
@puja108 I see nginx upstream changelog link is not in the announcement, but it is in the linked release notes https://github.com/giantswarm/releases/blob/master/kvm/v12.0.1

It would be redundant to repeat release notes, and announcements never had this level of detail to keep announcement message within reasonable readable size, especially now the announcements are aggregate of multiple announcements. As side-effect, two clicks are needed, to release notes and from there to upstream app changelog.

Second click to upstream changelog IMO is also not always necessary, if something like breaking change is important/relevant we'd stress it out in release notes. Still it is highly recommended to read linked upstream release notes since we don't have full picture/context of customer usage so extra check may reduce issues, and they could be informed about new features or improvements upstream which we don't repeat in release notes.

MartaD:
The downside of linking the upstream directly in the announcements might be that customers omit our release notes. This way they have to go through our note, which contains all the warnings/issues/requests and it’s more safe.

PR #365, opened 2020-07-21
AWS release v12.0.0

calvix:
towards https://github.com/giantswarm/giantswarm/issues/8560

## Changes compared to v11.5.0

- Kubernetes 1.17.9
- calico 3.15.1
- cluster-autoscaler 1.17.3
- aws-operator 8.7.5
- cluster-operator 2.3.2
- chart-operator 0.13.2
- app-operator 1.1.9
- metrics-server 1.1.1
- kiam 1.3.1
- kube-state-metrics 1.1.1
- node-exporter 1.3.0


aws/v12.0.0/release.yaml
```diff
@@ -0,0 +1,60 @@
...
+    version: 1.2.0
+  components:
+  - name: app-operator
+    version: 1.0.0
+  - name: aws-cni
+    version: 1.6.0
+  - name: aws-operator
+    version: 8.7.3
+  - name: calico
+    version: 3.12.3
```
tfussell, line 48: Can we bump this to 3.13+ for https://github.com/projectcalico/kube-controllers/pull/459?

calvix: I will be bumping many version here, this is still WIP :)

calvix: what about 3.14 and 3.15 ? should we try that as well?

tfussell: Cool, thanks. Just didn't want it to get lost. I'll add it to the releases board as well. We're using 3.14 in KVM 12.0 so I think that's safe. There could be some changes required for 3.15 or maybe it will "just work".

aws/v12.0.0/release.yaml
```diff
@@ -0,0 +1,60 @@
...
+    name: metrics-server
+    version: 1.1.1
+  - name: net-exporter
+    version: 1.9.0
+  - componentVersion: 1.0.1
+    name: node-exporter
+    version: 1.3.0
+  components:
+  - name: app-operator
+    version: 1.1.9
```
rossf7, line 42:
```suggestion
    version: 1.0.0
```

Sadly the versioning here is weird.

For Helm 2 we need the version to be 1.0.0. This is processed by app-operator 1.09 in the app collection. e.g. https://github.com/giantswarm/aws-app-collection/blob/0ab639b019a93e6b9fc8548b92b9980ac6b1daaf/helm/aws-app-collection-chart/templates/app-operator-unique-1.0.9.yaml

This is one of the things we will fix with Helm 3 for tenant clusters. The version will be 2.X and there will be an app-operator instance per version like cluster-operator.

cc @njuettner

PR #373, opened 2020-07-24
Check that releases satisfy requested component versions

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/11900

Adds a test which compares each active release against a provider-level `requests` file.
If an active release does not satisfy the requests or have an exception, the test fails.

The release names and component versions are both `semver.Constraints`, meaning they can match exact versions, version ranges, and certain wildcards.

For example, running with this `aws/requests.yaml` file:
```
releases:
    - name: 9.*
      requests:
        - name: calico
          version: 3.10.4
        - name: containerlinux
          version: 2512.2.1
          except:
            - releaseVersion: 9.3.5
              reason: too much aws-operator work needed
    - name: 11.5.0
      requests:
        - name: calico
          version: ">= 3.11.0"
        - name: kubernetes
          version: ">= 1.16.14"
```

yields the following output:
```
$ go test ./...
--- FAIL: Test_Releases (0.09s)
    --- FAIL: Test_Releases/case_1:_aws_releases_are_valid (0.07s)
        releases_test.go:341: Release v11.5.0 does not meet the requested version requirements:
            requested: calico: >= 3.11.0 	actual: 3.10.4,
            requested: kubernetes: >= 1.16.14 	actual: 1.16.13
        releases_test.go:341: Release v9.0.8 does not meet the requested version requirements:
            requested: calico: 3.10.4 	actual: 3.9.6
FAIL
FAIL	github.com/giantswarm/releases	0.103s
FAIL
```


go.mod
```diff
@@ -3,6 +3,7 @@ module github.com/giantswarm/releases
 go 1.13
 
 require (
+	github.com/Masterminds/semver v1.5.0
```
MarcelMue, line 6: Are we certain that this is the semver lib to use? It feels like we use different ones everywhere :(

releases_test.go
```diff
@@ -22,6 +23,86 @@ type kustomizationFile struct {
...
+		match, err := versionMatches(release, request.Name)
+		if err != nil {
+			return nil, microerror.Mask(err)
+		}
+
+		if match {
+			for _, component := range request.Requests {
+				releaseIsExcluded := false
+				if component.Exceptions != nil {
+					// Check the excluded releases for this component to see if our release is there
```
MarcelMue, line 88: Let's add `.` at the end of all comments at least kind of consistently :)

releases_test.go
```diff
@@ -202,6 +315,33 @@ func Test_Releases(t *testing.T) {
 				if !strings.Contains(readmeContent, fmt.Sprintf("https://github.com/giantswarm/releases/tree/master/%s/%s", tc.provider, release.Name)) {
 					t.Errorf("expected link in README.md to %s release %s", tc.provider, release.Name)
 				}
+
+				// Check that all active releases contain all requested component versions
+				if release.Spec.State == "active" {
```
MarcelMue, line 320: This will cause issues if there are legacy releases which are active (e.g. 9.0.X) but you are actually just trying to create a new 11.X.X release. Not sure how to solve this here though.

releases_test.go
```diff
@@ -135,6 +232,22 @@ func Test_Releases(t *testing.T) {
 				}
 			}
 
+			providerRequests := []releaseRequest{}
+			{
+				var providerRequestsFile requestsFile
+				providerRequestsData, err := ioutil.ReadFile(filepath.Join(tc.provider, "requests.yaml"))
```
MarcelMue, line 238: Could we put some of these into like a `key.go`. 
I know you didn't introduce this, but it feels like we are spreading magic strings like `requests.yaml` all over the place.

releases_test.go
stone-z: If your pattern matches both release 9.x and 11.x it's a dumb constraint and CI would have already complained when you added the request in the PR :smile:

go.mod
stone-z: This is in a lot of other repos :man_shrugging:

releases_test.go
stone-z: Just for you

releases_test.go
stone-z: That's a pet peeve of mine too. Moved a bunch of the filenames out

tfussell: I wrote that code in a hurry. Sorry, thanks for cleaning up.

go.mod
tfussell: This should be `github.com/Masterminds/semver/v3 v3.1.0` since v1 and v2 aren't under active development and aren't (fully) go module compatible AFAIK. See https://github.com/Masterminds/semver#package-versions We should upgrade this or notify owners if we see it in other repos too.

stone-z: Ah yeah, good catch. I always forget I have to version the import statements themselves

test_key.go
```diff
@@ -0,0 +1,8 @@
+package main
```
MarcelMue, line 1: Just got for `key.go` for the filename IMO

PR #377, opened 2020-07-28
Request Calico 3.13.4 for next 11.x release on all providers

stone-z:
Toward https://github.com/giantswarm/giantswarm/issues/11307 and https://github.com/giantswarm/giantswarm/issues/12334

Next 11.x releases should include Calico 3.13.4.

Should any other platform versions be bumped as well?

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -1,6 +1,6 @@
 releases:
-    - name: "> 11.5.0, < 12.0.0"
+    - name: "> 12.0.0, < 13.0.0"
       requests:
         - name: calico
-          version: ">= 3.13.4"
+          version: ">= 3.15.1"
           issue: https://github.com/giantswarm/giantswarm/issues/11307
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -1,0 +1,6 @@
+releases:
+    - name: "> 11.5.0, < 12.0.0"
+      requests:
+        - name: calico
+          version: ">= 3.13.4"
+          issue: https://github.com/giantswarm/giantswarm/issues/11307
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,0 +1,6 @@
+releases:
+    - name: "> 11.4.0, < 12.0.0"
+      requests:
+        - name: calico
+          version: ">= 3.13.4"
+          issue: https://github.com/giantswarm/giantswarm/issues/11307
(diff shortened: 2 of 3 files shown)


aws/requests.yaml
```diff
@@ -0,0 +1,6 @@
+releases:
+    - name: "> 11.5.0, < 12.0.0"
```
MarcelMue, line 2: Why not make this >12.0.0?
From what I understood here: https://github.com/giantswarm/giantswarm/issues/12334#issuecomment-664179537
The PM is not impacted by this version anymore as AWS CNI deals with it. We should still bump calico but it does not really need to be higher than 3.10.X in the 11.X.X releases IMO.

kvm/requests.yaml
```diff
@@ -0,0 +1,6 @@
+releases:
+    - name: "> 11.3.2, < 12.0.0"
```
MarcelMue, line 2: This one makes sense to me because the reported PM can actually happen as calico does IPAM calculations.

azure/requests.yaml
```diff
@@ -0,0 +1,6 @@
+releases:
+    - name: "> 11.4.0, < 12.0.0"
```
MarcelMue, line 2: I am not sure how much we already do with azure CNI here. I don't think calico does any IPAM calculations here.

stone-z:
I realize on some providers calico isn't managing much, but I'd like to still keep it updated as there are security and performance fixes regardless. It's difficult to tell what exactly it's doing on each provider by looking at the versions here. If it isn't doing anything, then it shouldn't be a disruptive update. But if someone is very strongly against updating it, they can exclude their release from the validation

calvix:
just FYI this is valid only for KVM or legacy AWS, as on Azure and 10.x.x AWS we don't use calico as CNI so the networking stack is not used as well as Kube controllers are not deployed.

aws/requests.yaml
stone-z: I thought it would be preferable to simply align the versions, but I can add a separate rule for 12+ and keep 11.x at 3.10

MarcelMue: Yes - I would prefer to align on 12+. I think that's reasonable and should still cover us very soon as all teams have / are working on 12.X releases.

stone-z: Ok, AWS 11.5.0 already includes 3.10.4, so I'll remove the request for anything on 11.x. 
The AWS 12.0.0 PR includes calico 3.15.1, so I'll request baselining the next 12.x for KVM and Azure on that as well. Then 12.x will cover everything we currently have open for calico.

PR #388, opened 2020-08-04
Archive AWS release v9.3.5

marians:
Unused and obsolete patch release


calvix:
I wonder when you archive releases, do you also delete the unused operators from aws-app-collection?

If no, we might wanna collect the recently archived releases and check if we can remove operators to free resources on CP

xh3b4sd:
Isn't this what `release-operator` is doing once it is active? /cc @yulianedyalkova

yulianedyalkova:
Yes, it's already active on all installations and ensures that you only have the operators referenced by a release running on the CP. So once you archive a release, it will clean up all the operators not being referenced by other releases.

PR #390, opened 2020-08-04
Archive unused azure releases

MartaD:
8.4.1 & 11.3.2 can be archived as no customer is using them.


whites11:
https://github.com/giantswarm/releases/pull/391

PR #392, opened 2020-08-04
Clear up releases list

MartaD:
I have been using the releases list recently to find out which patch releases are needed for CVE mitigation and it was hard to get this information from here.

My suggestion is to:
- remove archived releases from the list - it's growing fast and readability decreases 
- clearly mark deprecated - to make it easier for us to know which releases are deprecated and maybe it will be an additional push for customers to move 

In future, we could also mark EoL-ed releases here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -6,6 +6,7 @@ This repository contains Giant Swarm releases and changelogs. Releases can be in
 different states, namely `active`, `deprecated` and `wip`. With pull requests
 merged to the `master` branch releases get automatically deployed to all Giant
 Swarm control planes.
+Deprecated and unused releases are removed from this list and stored in Archived folder separate for each provider.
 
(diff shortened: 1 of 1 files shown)


tfussell:
This makes sense to me. The README was getting hard to read. I will need to update the validation logic for this to work with CircleCI--I'll try to do that today for you.

MartaD:
Thank you Thomas, doesn’t have to be today, I wasn’t aware of other implications related to this change. Let’s discuss today on standup, I’m ok with parking this until next week, no rush.

tfussell:
Finally got this fixed for you, Marta. Please take a look and make sure I didn't mess up what you had in mind.

tfussell:
@MartaD Do you still think this is relevant?

PR #393, opened 2020-08-04
AWS v11.5.1

njuettner:
Issue: https://github.com/giantswarm/giantswarm/issues/12584

### DIFF
--- a/aws/v11.5.1/release.diff
+++ b/aws/v11.5.1/release.diff
@@ -60,4 +60,4 @@ spec:                                                              spec:
   - name: kubernetes                                                 - name: kubernetes
     version: 1.16.13                                                   version: 1.16.13
   date: "2020-07-17T10:00:00Z"                                  |    date: "2020-08-04T12:34:57Z"
-  state: active                                                      state: active
+  state: deprecated                                             |    state: active
--- a/aws/v11.5.1/release.diff
+++ b/aws/v11.5.1/release.diff
@@ -46,7 +46,7 @@ spec:                                                              spec:
     releaseOperatorDeploy: true                                        releaseOperatorDeploy: true
     version: 8.7.3                                              <      version: 8.7.5
   - name: calico                                                     - name: calico
-    version: 3.10.4                                                    version: 3.10.4
+    version: 3.10.4                                             <      version: 3.15.1
   - name: cert-operator                                              - name: cert-operator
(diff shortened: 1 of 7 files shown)


aws/v11.5.1/release.diff
```diff
@@ -0,0 +1,63 @@
...
+    name: node-exporter                                                name: node-exporter
+    version: 1.2.0                                                     version: 1.2.0
+  components:                                                        components:
+  - name: app-operator                                               - name: app-operator
+    version: 1.0.0                                                     version: 1.0.0
+  - name: aws-cni                                                    - name: aws-cni
+    version: 1.6.0                                                     version: 1.6.0
+  - name: aws-operator                                               - name: aws-operator
+    releaseOperatorDeploy: true                                        releaseOperatorDeploy: true
+    version: 8.7.3                                              <      version: 8.7.5
```
xh3b4sd, line 47: There are 2 versions but only one change. How does that work? 🙈

tfussell: You mean the changelog doesn't show the changes from 8.7.4? I didn't get a chance to implement merging changelogs for intermediate releases yet, but I would like to do it soon.

xh3b4sd: In this case I think we just skipped one version but I was wondering why and if we are not missing something in the changelogs. The diff is fine for me.

njuettner: dependency for aws-operator (k8scloudconfig) with the correct calc for max pods was added in 8.7.5 😞 and not 8.7.4

xh3b4sd: Still my question is not answered, what happened in the version before?

xh3b4sd:
@giantswarm/team-firecracker let's get this out.

aws/v11.5.1/release.diff
marians: The aws-operator changelog shows:

- 8.7.5 changed: "Adjust number of host network pods on worker node for aws-cni"
- 8.7.4 fixed: "Adjust MAX_PODS for master and worker nodes to max IP's per ENI when using aws-cni"

I consider the two item as closely related and in my release notes curation work decided to summarize it all as

> Adjust `MAX_PODS` for master and worker nodes to max IPs per Elastic Network Interface (ENI)

If we lose important information here, please let me know.

PR #424, opened 2020-08-24
Add release management functions

tfussell:
As recently discussed in SIG Operator, this PR adds the concept of a release patch. A patch is applied to a Release CR to create a new updated Release CR. This allows us to replace the usage of many `--component` flags in `opsctl release` and `devctl release`. Additionally, the patch can be saved in the release so the release can be verified with CI by release PR reviewers (replaces `release.diff`). Here is an example of the patch used to create the upcoming AWS 9.3.8 release:
```yaml
version: v9.3.8
date: "2020-08-28T15:00:00Z"
components:
  - name: aws-operator
    reference: update-k8scc-legacy-alt
  - name: kubernetes
    version: 1.16.14
  - name: calico
    version: 3.10.4
  - name: etcd
    version: 3.3.25
  - name: containerlinux
    version: 2512.3.0
```
It feels a little strange putting this in the `releases` repo, but it also makes sense in some ways. Should it maybe go somewhere else?


tfussell:
Towards https://github.com/giantswarm/giantswarm/issues/12946

tfussell:
Closing in favor of https://github.com/giantswarm/releaseclient/pull/1

PR #432, opened 2020-08-31
Azure 12.1.0 version for helm 3 upgrade

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/10751

PR #437, opened 2020-09-02
KVM 12.3.0 version for helm 3 upgrade

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13095


tomahawk28:
the announcement came from https://github.com/giantswarm/releases/pull/432 mostly

rossf7:
@tomahawk28 This LGTM and we need it so we can preview Helm 3 with customers.

But we're not planning to merge till w/c 14th September. So let's make it draft so that's clear. Please also include that in the PR desc.

tomahawk28:
I don't think keep open this until the middle of this month makes sense, I will close it open another one at that point.

PR #442, opened 2020-09-10
Request Flatcar 2512.4.0 in all next releases

stone-z:
cc @giantswarm/sig-product please update Flatcar in next-out releases

Issue https://github.com/giantswarm/giantswarm/issues/13150

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -4,3 +4,18 @@ releases:
         - name: calico
           version: ">= 3.15.1"
           issue: https://github.com/giantswarm/giantswarm/issues/11307
+    - name: "> 12.1.4, < 13.0.0"
+      requests:
+        - name: flatcar
+          version: ">= 2512.4.0"
+          issue: https://github.com/giantswarm/giantswarm/issues/13150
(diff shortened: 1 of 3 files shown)


tfussell (approved):
It's too bad the rules have to be so complicated, but it otherwise looks good.

stone-z:
> It's too bad the rules have to be so complicated, but it otherwise looks good.

Agreed. I'm not sure how to simplify the base rules since we release new versions out of chronological order (e.g. a 9.x between two 12.xs). I was thinking I could introduce a combination rule which just specifies many names for one request. Would be open to suggestions

I put a note on my board to automate creating and removing them as well, this isn't super fun to do manually

stone-z:
> The basic idea is that all new releases should match this request, right? Unless we have some old release that can't be updated, I wonder if we could base the restriction off of the release date (i.e. releases created after today, September 10th) rather than matching version ranges directly.

I had considered that as well -- it could help as an additional option for cases like this (Flatcar) where it applies to everything, but in other cases we'd still need to be able to bump within version ranges (like `calico`). The part I don't like is the validation then depends on when you run it, which feels a Bad Thing (tm)

PR #466, opened 2020-10-05
Add cert-manager-app v2.3.0 request

corest:


cokiengchiara:
thanks @corest ! I am not familiar with how to review this kind of PR. 🙈 

FYI @giantswarm/team-halo , @MartaD explained this is the new "request changes in next platform releases" feature. Now anyone creating AWS release for example 12.3.1 will have to include cert-manager 2.3.0 or his Circle CI build will fail. This is how we can handoff a feature for provider team to release (if not critical to release right now).

aws/requests.yaml
```diff
@@ -1,4 +1,9 @@
 releases:
+    - name: "> 12.3.0, < 13.0.0"
```
cokiengchiara, line 2: So this includes cert manager v2.3.0 in **all** AWS releases between (but not including) 12.3.0 and 13.0.0, right?

Just curious, what happens when there's a new cert manager v2.4.0? Can we override the v2.3.0 that's already released?

cokiengchiara (approved):
LGTM based on what's described here https://intranet.giantswarm.io/docs/product/releases/requesting-changes-in-next-platform-release/#example but I am new to this

aws/requests.yaml
MartaD: This means it can be 2.3.0 or newer but it can stay as 2.3.0. If you want to enforce 2.4.0 you edit this request from 2.3.0 to 2.4.0 and all new releases (created after your edit) need to include 2.4.0

cokiengchiara: cool! 😎 thanks!

cokiengchiara:
hi @corest will you merge this? I'm relying on your PR so Team Halo no longer needs to make a platform release to make cert manager v2.3.0 available :D

PR #476, opened 2020-10-13
Added requests for next azure release

whites11:
Enforce kubernetes 1.17.12, external-dns 1.5.0 and newest flatcar for next azure release

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -14,9 +14,6 @@ releases:
         - name: containerlinux
           version: ">= 2512.4.0"
           issue: https://github.com/giantswarm/giantswarm/issues/13150
-        - name: external-dns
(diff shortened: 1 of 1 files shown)


azure/requests.yaml
```diff
@@ -14,9 +14,6 @@ releases:
         - name: containerlinux
           version: ">= 2512.4.0"
           issue: https://github.com/giantswarm/giantswarm/issues/13150
-        - name: external-dns
```
fiunchinho, line 17: Not sure how this works so bear with me. Why is this removed?

whites11: It was added in the wrong section, for clusters < 13.0.0.
We're not going to release any new release < 13.0.0

tuommaki (approved):

PR #478, opened 2020-10-14
Update requests.yaml

MartaD:
Added app-operator and chart-operator to releases 13.0.0+ as per @rossf7 request. Removed duplicate request for containerLinux in versions >11.4.0 <13.0.0


whites11 (approved):

azure/requests.yaml
```diff
@@ -34,3 +24,7 @@ releases:
           issue: https://github.com/giantswarm/giantswarm/issues/13671
         - name: kubernetes
           version: ">= 1.17.12"
+        - name: app-operator
+          version: ">=2.3.2"
+        - name: chart-operator
+          version: ">=2.3.3"
```
rossf7, line 30:
```suggestion
          version: ">=2.3.5"
```

Thanks Marta! Sorry to be a pain but could you include the chart-operator v2.3.5? It's a bug fix release we created yesterday.

https://github.com/giantswarm/chart-operator/blob/master/CHANGELOG.md#235---2020-10-13

MartaD: sure, we're not testing the release yet anyway

rossf7 (approved):
Thanks, LGTM

azure/requests.yaml
```diff
@@ -34,3 +24,7 @@ releases:
           issue: https://github.com/giantswarm/giantswarm/issues/13671
         - name: kubernetes
```
MartaD, line 25: @whites11 you wanted 1.17.12 for the alpha release right? is this request covering alpha as well? For >=13.0.0 we want 1.18.x. And if it covers alphas then this PR would create a mess..

azure/requests.yaml
```diff
@@ -34,3 +24,7 @@ releases:
...
           version: ">= 1.17.12"
```
MartaD, line 26: @whites11 you wanted 1.17.12 for the alpha release right? is this request covering alpha as well? For >=13.0.0 we want 1.18.x. And if it covers alphas then this PR would create a mess..

PR #483, opened 2020-10-16
Update requests.yaml

MartaD:
Again - adding chart-operator and app-operator requests to Azure v13.0.0

My previous PR was open for too long and it's easier to recreate it. Ross please review again, sorry for inconvenience.

Chiara I saw you have a PR for this open, but you're adding it to version <13.0.0 and we don't plan any releases for that version. Please check if this is the right one, those apps will be updated in Azure v13.0.0.

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -33,6 +33,7 @@ releases:
         - name: kubernetes
           version: ">= 1.17.13"
         - name: app-operator
-          version: ">= 2.3.2"
+          version: ">= 2.3.3"
+          issue: https://github.com/giantswarm/giantswarm/issues/13648
         - name: chart-operator
           version: ">= 2.3.5"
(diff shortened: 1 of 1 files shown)


azure/requests.yaml
```diff
@@ -34,3 +32,7 @@ releases:
           issue: https://github.com/giantswarm/giantswarm/issues/13671
         - name: kubernetes
           version: ">= 1.17.13"
+        - name: app-operator
+          version: ">= 2.3.2"
```
rossf7, line 36:
```suggestion
          version: ">= 2.3.3"
```

We shipped 2.3.3 yesterday which has a fix for this PM. https://github.com/giantswarm/giantswarm/issues/13648

Our next release should only affect control planes. So I hope we don't need to update this again.

rossf7 (approved):
Thanks @MartaD 

LGTM once the app-operator version is bumped. 

>  but you're adding it to version <13.0.0 and we don't plan any releases for that version.

Got it. We may need to do a patch version with the fix for this PM. Still discussing in batman.

azure/requests.yaml
MartaD: Thanks, I added it together with the link to PM.

PR #484, opened 2020-10-16
hotfix; KVM 12.3.0 upgarding app-operator into 2.3.4

tomahawk28:
We just release a new app-operator with hotfix 2.3.4, and we don't introduce 12.3.0 to any customers yet.

So I think this is safe to merge.


tomahawk28:
going with a ping

PR #493, opened 2020-10-21
Azure 12.1.1 with cluster-operator patch version

tomahawk28:
Toward https://github.com/giantswarm/giantswarm/issues/13869


azure/v12.1.1/announcement.md
```diff
@@ -0,0 +1,3 @@
+**Giant Swarm release v12.1.1 for Azure is available**. This patch release fixes a problem causing the accidental deletion and reinstallation of Preinstalled Apps (such as CoreDNS) in 9.x.x tenant clusters. Please upgrade all older clusters to this version in order to prevent possible downtime.
```
rossf7, line 1:
```suggestion
**Giant Swarm release v12.1.1 for Azure is available**. This patch release fixes a problem causing the accidental deletion and reinstallation of Preinstalled Apps (such as CoreDNS). Please upgrade all older clusters to this version in order to prevent possible downtime. 
```

I think we just need to remove the 9.x.x. For Azure only the alpha uses node pools.

rossf7 (approved):
LGTM

cokiengchiara:
Giving it until 16:00 in sig-product for feedback but it's the same as AWS so shouldn't be controversy.

cokiengchiara (approved):
LGTM. Clarified w Ross and Jihyuk that notes to SEs are still relevant.

PR #500, opened 2020-10-28
Add announcement to AWS v12.5.2

cokiengchiara:
Created announcement. Trying to communciate that:

- Small change, so pls try to upgrade
- Low blast radius, since likely not to delete cluster
- but in case you do, and app operator crashes, it's not processing other clusters


tomahawk28 (approved):
LGTM :)

PR #518, opened 2020-11-12
Add required status fields to KVM release CRs

rossf7:
releases-kvm-unique is failed on dragon due to missing required fields in the release CR.

```
kg get app releases-kvm-unique -o yaml | yq .status
{
  "appVersion": "",
  "release": {
    "lastDeployed": null,
    "reason": "Upgrade \"releases-kvm-unique\" failed: cannot patch \"v11.2.0\" with kind Release: Release.release.giantswarm.io \"v11.2.0\" is invalid: status.inUse: Required value && cannot patch \"v11.2.1\" with kind Release: Release.release.giantswarm.io 
...
```

This fixes it by adding the missing fields as suggested by @tfussell in https://gigantic.slack.com/archives/C03CPNRTS/p1605170707041300?thread_ts=1605170388.040800&cid=C03CPNRTS

### DIFF
--- a/kvm/v11.2.0/release.yaml
+++ b/kvm/v11.2.0/release.yaml
@@ -52,3 +52,6 @@ spec:
     version: 3.10.0
   date: "2020-02-26T12:00:00Z"
   state: deprecated
+status:
+  inUse: false
+  ready: false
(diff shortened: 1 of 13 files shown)


rossf7:
Tested on `geckon`.

tfussell (approved):
LGTM. Eventually we should make these optional but this is a good solution for now.

PR #532, opened 2020-11-23
Request: update cert-manager for AWS releases

glitchcrab:
towards giantswarm/giantswarm/issues/14281

This release allows cert-manager components to be scheduled on master nodes, so the deployment won't fail when a cluster is created without any nodepools.

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -1,5 +1,5 @@
 releases:
-    - name: "> 12.7.0, < 13.0.0"
+    - name: "> 13.0.0, < 14.0.0"
       requests:
         - name: cert-manager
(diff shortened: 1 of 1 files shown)


aws/requests.yaml
```diff
@@ -1,4 +1,9 @@
 releases:
+    - name: "> 12.7.0, < 13.0.0"
```
calvix, line 2: we alre already drafting 13.0.0 so does this make sense?

there is no planned release 12.8.0 for now

glitchcrab: ok, bumped it to 13.0.0

calvix (approved):

PR #533, opened 2020-11-24
Introduce faster app CRs status change to the next AWS release

tomahawk28:
To bring https://github.com/giantswarm/roadmap/issues/191 into TCs. 

Related slack announcement (when we deploy this to CPs) : https://gigantic.slack.com/archives/C04TGHDEF/p1604052483005300


azure/requests.yaml
```diff
@@ -1,4 +1,12 @@
 releases:
+    - name: "> 13.0.0-beta1"
```
tomahawk28, line 2: I saw there is a beta version on-going, should I change it to `13.0.0` as normal?

rossf7: AIUI the beta release is a preview release for node pools on Azure.

I'm not sure. I think it's best to ask Celestial.

rossf7 (commented):
I think we should get approvals from the provider teams as well as batman. So they know we want to change something.

I think with this and the automated Prow tests in https://github.com/giantswarm/giantswarm/issues/13971 it will improve the current situation.

It might also be better to do PRs per provider. That adds a bit of overhead but the PRs are small and I think it makes the review easier.

@cokiengchiara @tomahawk28 WDYT?

tomahawk28:
@rossf7 Sure I will change this PR as AWS release only and create additional PRs for other providers.

tomahawk28:
Asked reviews from Firecracker. 🧨

njuettner (approved):

njuettner (approved):

njuettner (approved):

njuettner (approved):

cokiengchiara (approved):
LGTM, thanks @tomahawk28 !

PR #535, opened 2020-11-24
Introduce faster app CRs status change to the next Azure release

tomahawk28:
To bring giantswarm/roadmap#191 into TCs.

Related slack announcement (when we deploy this to CPs) : https://gigantic.slack.com/archives/C04TGHDEF/p1604052483005300


azure/requests.yaml
```diff
@@ -1,4 +1,12 @@
 releases:
+    - name: "> 13.0.0-beta1"
```
tomahawk28, line 2: @giantswarm/team-celestial Is this okay to set the beta version as a threshold?

whites11 (approved):

cokiengchiara (approved):
LGTM since @whites11 is ok with 
`    - name: "> 13.0.0-beta1"`

azure/requests.yaml
whites11: I don't see why not

tomahawk28: Okay, let's go with this!

PR #540, opened 2020-11-26
Deprecate v12.0.2

MartaD:
Deprecate v12.0.2 because v12.1.2 has the same k8s version


whites11 (approved):
Do we want to remove it completely maybe?
Otherwise LGTM

MartaD:
There are 6 clusters running this version, can't remove yet.

whites11:
> There are 6 clusters running this version, can't remove yet.

Good, go for it then

PR #543, opened 2020-11-27
Add requests for upgrade issues

MarcelMue:

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -1,6 +1,12 @@
 releases:
     - name: "> 13.0.0, < 14.0.0"
       requests:
+        - name: node-exporter
+          version: ">= 1.7.0"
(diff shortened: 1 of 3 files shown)


whites11 (approved):

aws/requests.yaml
```diff
@@ -1,6 +1,12 @@
 releases:
     - name: "> 13.0.0, < 14.0.0"
```
whites11, line 2: shouldn't it be >= 13.0.0 ?

tuommaki (approved):
Sure!

aws/requests.yaml
MarcelMue: Fixed.

stone-z (approved):
LGTM

PR #557, opened 2020-12-09
Request chart-operator v2.5.1 for Azure 12.X

rossf7:
chart-operator v2.5.1 is already in the KVM 13.0.1 release.

Also adding to requests in case we need to do patch releases.


fiunchinho (approved):
LGTM

tomahawk28 (approved):

PR #592, opened 2021-01-26
Switched to a different semver library

whites11:
We currently use the `github.com/Masterminds/semver/v3` to ensure requests are met for all providers.

That library has a problem with `-dev` releases though:

```
5.2.2-dev >= 5.2.0 ? false
5.2.0-dev >= 5.2.0 ? false
5.3.0-dev >= 5.2.0 ? false
```

This PR switches to the `github.com/blang/semver/v4` library that behaves correctly:

```
5.2.2-dev >= 5.2.0 ? true
5.2.0-dev >= 5.2.0 ? false
5.3.0-dev >= 5.2.0 ? true
```

Unluckily this requires a slightly different syntax in the range specification.

WDYT?

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/giantswarm/releases
 go 1.13
 
 require (
-	github.com/blang/semver v3.5.0+incompatible
+	github.com/blang/semver/v4 v4.0.0
 	github.com/giantswarm/apiextensions v0.4.14
 	github.com/giantswarm/microerror v0.2.0
 	github.com/giantswarm/versionbundle v0.2.0
--- a/go.mod
+++ b/go.mod
@@ -3,7 +3,7 @@ module github.com/giantswarm/releases
 go 1.13
 
 require (
-	github.com/Masterminds/semver/v3 v3.1.0
+	github.com/blang/semver v3.5.0+incompatible
 	github.com/giantswarm/apiextensions v0.4.14
 	github.com/giantswarm/microerror v0.2.0
 	github.com/giantswarm/versionbundle v0.2.0
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -1,10 +1,10 @@
 releases:
-    - name: "> 13.1.0, < 14.0.0"
+    - name: "> 13.1.0 < 14.0.0"
       requests:
         - name: cert-manager
           version: ">= 2.4.1"
           issue: https://github.com/giantswarm/giantswarm/issues/14706
-    - name: "> 13.0.0, < 14.0.0"
+    - name: "> 13.0.0 < 14.0.0"
       requests:
         - name: app-operator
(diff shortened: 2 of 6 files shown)


axbarsan (approved):
It might be worth adding this to `fmt`, like we do with the `yaml` package

whites11:
> It might be worth adding this to `fmt`, like we do with the `yaml` package

I don't get you, what do you mean?

axbarsan:
> I don't get you, what do you mean?

Check this out
https://github.com/giantswarm/fmt/blob/master/go/yaml_marshalling.md

go.mod
```diff
@@ -3,7 +3,7 @@ module github.com/giantswarm/releases
 go 1.13
 
 require (
-	github.com/Masterminds/semver/v3 v3.1.0
+	github.com/blang/semver v3.5.0+incompatible
```
stone-z, line 6: Can we use the latest? It is compatible with go modules.
```suggestion
	github.com/blang/semver/v4 v4.0.0
```

fiunchinho (changes requested):
We need to replace the library because it had a bug in some specific case. Can you add that case as a test?

stone-z (approved):
So in general this is ok with me as long as it preserves the behavior. What is the use case for requesting things in `-dev` releases?

We use `masterminds` a bunch of other places, so it would be nice to be consistent, but if the `-dev` requests are needed, please also update [the doc](https://intranet.giantswarm.io/docs/product/releases/requesting-changes-in-next-platform-release/) with examples of the new format. The old behavior is also described in the text itself

whites11:
> We need to replace the library because it had a bug in some specific case. Can you add that case as a test?

the test of a test?
The only go code we have here is actually a go test to ensure validation is ensured.
How do you suggest to do it? Just adding a new test function?

fiunchinho:
> > We need to replace the library because it had a bug in some specific case. Can you add that case as a test?
> 
> the test of a test?
> The only go code we have here is actually a go test to ensure validation is ensured.
> How do you suggest to do it? Just adding a new test function?

I noticed we are only testing our own requests, so I guess it doesn't make sense to test other things. Nevermind then.

go.mod
whites11: The use case is while crafting a new release while not having a final release of the provider operator yet.
For example, we are currently crafting release `14.0.0` for azure, that has a request for azure operator > 5.2.0.
We set a reference to the latest master in the release PR for now, and the version is 5.2.2-dev.
Does it make sense?

stone-z: Ok that makes more sense. I was trying to understand why we'd ever request a dev version, now I get that it's the version which is being _compared to_ the request.

stone-z: That comparison behavior is probably incorrect anyway though -- maybe you should make an issue with `masterminds` as well

PR #599, opened 2021-02-02
WIP: Create release 13.1.0

paurosello:


aws/v13.1.0/release.yaml
```diff
@@ -0,0 +1,69 @@
...
+    version: 1.2.1
+  - name: net-exporter
+    version: 1.9.2
+  - componentVersion: 1.0.1
+    name: node-exporter
+    version: 1.7.1
+  components:
+  - name: app-operator
+    releaseOperatorDeploy: true
+    version: 3.1.0
```
rossf7, line 45:
```suggestion
    version: 3.2.0
```

@paurosello There is a problem with finalizers in this release that @xh3b4sd hit testing on ginger.

In 3.1.0 the finalizer got changed from `operatorkit.giantswarm.io/app-operator` to `operatorkit.giantswarm.io/app` which is wrong and causes problems on deletion.

In 3.2.0 we use the correct value which is `operatorkit.giantswarm.io/app-operator-app` and there is migration logic to use the new value.

I'll do a PR for requests but posting here so we don't hit this.





The finalizer got renamed from

aws/v13.1.0/release.yaml
```diff
@@ -0,0 +1,64 @@
...
+    version: 9.3.8
+  - name: calico
+    version: 3.15.3
+  - name: cert-operator
+    reference: 0.1.0-2
+    releaseOperatorDeploy: true
+    version: 0.1.0
+  - name: cluster-operator
+    releaseOperatorDeploy: true
+    version: 3.5.1
```
anvddriesch, line 54:
```suggestion
    version: 3.6.0
```
If it's okay, let's include the cluster status fix in the release.

stone-z:
Reminder: if you wind up bumping the `cluster-operator` version again for this release, you'll have to also bump `cert-operator` to 1.0.0 or later (which may be a good thing to do anyway)

alex-dabija (approved):

aws/v13.1.0/release.diff
```diff
@@ -0,0 +1,70 @@
...
+metadata:                                                          metadata:
+  annotations:                                                       annotations:
+    giantswarm.io/docs: https://docs.giantswarm.io/reference/cp        giantswarm.io/docs: https://docs.giantswarm.io/reference/cp
+  creationTimestamp: null                                            creationTimestamp: null
+  name: v13.0.0                                                 |    name: v13.1.0
+spec:                                                              spec:
+  apps:                                                              apps:
+  - name: cert-exporter                                              - name: cert-exporter
+    version: 1.3.0                                              |      version: 1.6.1
+  - componentVersion: 1.0.2                                     |    - name: cert-manager
```
calvix, line 14: why do we miss the `componentVersion` section in the new release?

aws/v13.1.0/release.diff
```diff
@@ -0,0 +1,70 @@
...
+    version: 2.3.3                                              <
+  - name: chart-operator                                             - name: chart-operator
+    version: 2.5.1                                              |      version: 2.12.0
+  - componentVersion: 1.18.1                                         - componentVersion: 1.18.1
+    name: cluster-autoscaler                                           name: cluster-autoscaler
+    version: 1.18.3                                                    version: 1.18.3
+  - componentVersion: 1.6.5                                          - componentVersion: 1.6.5
+    name: coredns                                                      name: coredns
+    version: 1.2.0                                                     version: 1.2.0
+  - componentVersion: 0.7.4                                     |    - name: external-dns
```
calvix, line 25: same here, missing componentVersion

calvix (approved):
i assume its tested, but the missing `componentVersion` should be in the new release

aws/v13.1.0/release.diff
alex-dabija: I added the missing componentVersion.

aws/v13.1.0/release.diff
alex-dabija: I added the missing componentVersion.

PR #608, opened 2021-02-08
request new external-dns to be added to azure release

glitchcrab:
towards giantswarm/giantswarm/issues/15573

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,5 +1,5 @@
 releases:
-    - name: "> 14.0.0"
+    - name: "> 14.1.0"
       requests:
         - name: external-dns
           version: ">= 2.1.0"
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,4 +1,9 @@
 releases:
+    - name: "> 14.0.0"
+      requests:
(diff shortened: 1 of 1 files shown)


azure/requests.yaml
```diff
@@ -1,4 +1,9 @@
 releases:
+    - name: "> 14.0.0"
```
whites11, line 2:
```suggestion
    - name: "> 14.1.0"
```

Tests for 14.1.0 are already done, so this will happen with the following release

glitchcrab: surely using the greater than operator `>` means that this logic is still correct? My changes are requested in release versions _after_ 14.0.0?

MartaD: but it will mess with releasing 14.1.0 and we'll have to add an exception and all that, so it would be easier to change it to 14.1.0

glitchcrab: ohhhh sorry i misread this. i thought Chris said 'Tests for 14.0.0 are already done' which is why i saw no issues with including it in 14.1.0 :facepalm: 

it's too early in the morning for me.

whites11 (approved):

PR #614, opened 2021-02-15
Request app-operator v3.2.0 for AWS >= 13.1.0

rossf7:
See https://github.com/giantswarm/giantswarm/issues/15392 


aws/requests.yaml
```diff
@@ -17,9 +21,6 @@ releases:
           issue: https://github.com/giantswarm/giantswarm/issues/15573
     - name: "> 13.0.0 < 14.0.0"
       requests:
-        - name: app-operator
-          version: ">= 3.2.0"
-          issue: https://github.com/giantswarm/giantswarm/issues/15392
```
rossf7, line 22: The < 14.0.0 is incorrect which is why we need another PR :(

cokiengchiara (approved):
thank you!

tomahawk28 (approved):

aws/requests.yaml
```diff
@@ -4,6 +4,10 @@ releases:
         - name: coredns
           version: ">= 1.4.0"
           issue: https://github.com/giantswarm/giantswarm/issues/13771
+    - name: ">= 13.1.0"
```
cokiengchiara, line 7: But why not `>13.0.0`? so it includes, say, 13.0.1?

cokiengchiara: wrong component

rossf7: > just wondering, why not >13.0.0? so it includes, say, 13.0.1?

I didn't think of that. The ranges are a bit tricky because 14.0.0 already exists but this is done now and CI is passing.

cokiengchiara: oki doki!

rossf7:
Going with this.

PR #615, opened 2021-02-15
Request app-operator 3.2.0 for Azure > 14.1.0

rossf7:
See giantswarm/giantswarm#15392

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -6,9 +6,6 @@ releases:
           issue: https://github.com/giantswarm/giantswarm/issues/13771
     - name: "> 14.1.0"
       requests:
-        - name: app-operator
-          version: ">= 3.2.0"
-          issue: https://github.com/giantswarm/giantswarm/issues/15392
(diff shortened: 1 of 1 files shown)


tomahawk28 (approved):

azure/requests.yaml
```diff
@@ -6,6 +6,9 @@ releases:
           issue: https://github.com/giantswarm/giantswarm/issues/13771
     - name: "> 14.1.0"
```
cokiengchiara, line 7: Why not > 14.0.0 so it includes 14.0.1? haha

whites11 (approved):

azure/requests.yaml
rossf7: Existing releases make this tricky. CI passes now and I think its better.

cokiengchiara: oki!

cokiengchiara (approved):
LGTM TY!

PR #616, opened 2021-02-15
Request app-operator 3.2.0 for KVM > 13.0.1

rossf7:
See giantswarm/giantswarm#15392

### DIFF
--- a/kvm/requests.yaml
+++ b/kvm/requests.yaml
@@ -6,9 +6,6 @@ releases:
           issue: https://github.com/giantswarm/giantswarm/issues/13771
     - name: "> 13.0.1"
       requests:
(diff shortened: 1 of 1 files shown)


kvm/requests.yaml
```diff
@@ -6,6 +6,9 @@ releases:
           issue: https://github.com/giantswarm/giantswarm/issues/13771
     - name: "> 13.0.1"
```
cokiengchiara, line 7: Why not > 13.0.0 so it includes 13.0.1?

cokiengchiara (approved):
LGTM

yulianedyalkova (approved):

PR #625, opened 2021-02-25
Define and activate release v12.4.0 for kvm

yulianedyalkova:

### DIFF
--- a/README.md
+++ b/README.md
@@ -187,6 +187,8 @@ to all Giant Swarm installations.
     - [v13.0.0](https://github.com/giantswarm/releases/tree/master/kvm/v13.0.0)
 
 - v12
+  - v12.4
+    - [v12.4.0](https://github.com/giantswarm/releases/tree/master/kvm/v12.4.0)
   - v12.3
     - [v12.3.3](https://github.com/giantswarm/releases/tree/master/kvm/v12.3.3)
     - [v12.3.2](https://github.com/giantswarm/releases/tree/master/kvm/v12.3.2)
--- a/kvm/kustomization.yaml
+++ b/kvm/kustomization.yaml
@@ -7,6 +7,7 @@ resources:
 - v12.3.1
 - v12.3.2
 - v12.3.3
+- v12.4.0
 - v13.0.0
 - v13.0.1
 transformers:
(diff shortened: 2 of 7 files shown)


cokiengchiara:
thanks @yulianedyalkova ! @giantswarm/team-batman-engineers , I'm still a bit confused w https://github.com/giantswarm/releases/blob/master/kvm/requests.yaml ✌️ 

Is our request here for:

* app-operator 2.7.0 or 3.2.0? 
* cluster-operator 0.24.2 or 0.23.22?

rossf7:
@yulianedyalkova @cokiengchiara Ideally we would go with

- app-operator 3.2.0
- chart-operator 2.9.0

For cluster-operator we need to check 0.24.2. There was a problem on AWS with generating the app-operator kubeconfig because the labels on the secrets created by cert-operator have changed. @stone-z Is that fixed in 0.24.2?

Yulia, I can help with testing this week if that is helpful? chart-operator has a bug fix for creating too many Helm secrets on error. So it would be great to get that in.

Chiara, this is why we need an automated Tekton test for pre-installed apps. To remove some of the toil from testing. I hope we can work on that in our next cycle. 🙏

stone-z:
> There was a problem on AWS with generating the app-operator kubeconfig because the labels on the secrets created by cert-operator have changed. @stone-z Is that fixed in 0.24.2

Yep, that's fixed by 0.24.2. (For posterity, the problem was that the `certs` library itself had dropped the old labels when it was updated for AWS HA masters, but this was never back-ported for legacy. AWS has been using the newer `certs` version but it was a problem on Azure and KVM, which were on a very old version).

yulianedyalkova:
It turns out that backporting our fix to `v13.x` is with higher priority than the current release so this PR might not be needed from our side for now. @rossf7 do I need to update something in that release before testing?

rossf7:
> It turns out that backporting our fix to v13.x is with higher priority than the current release so this PR might not be needed from our side for now. @rossf7 do I need to update something in that release before testing?

@yulianedyalkova Yes please use app-operator 3.2.0 and chart-operator 2.9.0. They both have fixes that we need.

PR #632, opened 2021-03-02
Add release with azure CAPZ controllers

MarcelMue:
towards https://github.com/giantswarm/giantswarm/issues/15996


MarcelMue:
No need to review but this PR now works to deploy our beautiful CAPI controllers, so feel free to add more of the other controllers (CAPZ here).

fiunchinho:
I only see `cluster-api-core` , but not `capz`. What am I missing?

MarcelMue:
> I only see `cluster-api-core` , but not `capz`. What am I missing?

Give us some time man :P 

This is only WIP, we will add all the controllers soon :)

azure/v20.0.0-v1alpha3/release.yaml
```diff
@@ -11,7 +11,8 @@ spec:
   - name: cert-exporter
     version: 1.6.0
   - name: chart-operator
-    version: 2.11.0
+    version: 2.11.0-a332e330b9a1e5a12e8484038f9120f243c578fe # https://github.com/giantswarm/chart-operator/pull/721
```
QuentinBisson, line 14: Could you update capv as well?

kubasobon: Sure.

whites11:
@MarcelMue we are taking over ownership of this PR

azure/v20.0.0/release.yaml
```diff
@@ -0,0 +1,80 @@
...
+    name: node-exporter
+    version: 1.7.1
+  - name: cluster-autoscaler
+    version: 1.19.1
+  - name: azure-scheduled-events
+    version: 0.2.2
+  components:
+  - name: app-operator
+    releaseOperatorDeploy: true
+    version: 4.2.0
```
rossf7, line 40:
```suggestion
    releaseOperatorDeploy: false
    version: 5.2.0
```

@MarcelMue Can we use the latest app-operator here?

There is a fix for the chart CR watcher https://github.com/giantswarm/app-operator/pull/790

Also releaseOperatorDeploy should be false because cluster-apps-operator will manage the app CR for the per cluster instance.

MarcelMue: Sure - sounds reasonable to me. Any celestial can hit the `commit` button when they do their next round of testing :)

azure/v20.0.0-alpha1/release.yaml
```diff
@@ -0,0 +1,81 @@
...
+  - name: kubernetes
+    version: 1.19.8
+  - name: containerlinux
+    version: 2605.12.0
+  - name: calico
+    version: 3.15.3
+  - name: etcd
+    version: 3.4.14
+  date: "2021-03-02T14:50:41Z"
+  state: wip
```
alex-dabija, line 78: the state of the release need to be changed to preview instead of wip.

alex-dabija, line 78: We should also set the notice field to let the customers know that this is just a test release and that there's no upgrade path to and from it.

fiunchinho: What notice should we use?

> This is a preview release only intended to be used for testing. No support is offered at the moment.

maybe?

alex-dabija (approved):

T-Kukawka (approved):

PR #633, opened 2021-03-04
Add release with AWS CAPA controllers

paurosello:
Towards https://github.com/giantswarm/giantswarm/issues/15996


aws/v20.0.0-v1alpha3/README.md
```diff
@@ -0,0 +1 @@
+# :zap: Giant Swarm Release v20.0.0-v1alpha3 for Azure :zap:
```
anvddriesch, line 1:
```suggestion
# :zap: Giant Swarm Release v20.0.0-v1alpha3 for AWS :zap:
```

aws/kustomization.yaml
```diff
@@ -16,5 +16,6 @@ resources:
 - v14.0.0
 - v14.1.1
 - v14.2.0
+- v20.0.0
```
calvix, line 19: should it be named something like `v20.0.0-alpha1`?

calvix (approved):

anvddriesch:
So what do we need to do to actually merge this?
This release really only matters for test installations and is not intended for customers. To create clusters we always need
- kyverno + newest policies
- all the CAPI CRDs
- CRD conversion to v1alpha3
- AWS credentials + IAM policies
Some of this we definitely don't have in all installations.

aws/kustomization.yaml
anvddriesch: We could do that. Right now I used this name since it is the one used in the `kyverno policies` https://github.com/giantswarm/kyverno-policies

calvix:
if its ready to merge let's do it :p now we really need it on the test installations only

PR #634, opened 2021-03-04
Add release with vSphere CAPV controllers

QuentinBisson:

### DIFF
--- a/helm/releases-vmware/Chart.yaml
+++ b/helm/releases-vmware/Chart.yaml
@@ -1,0 +1,6 @@
+apiVersion: v1
+name: releases-vmware
+version: [[ .Version ]]
+appVersion: master
+description: Giant Swarm releases for vmware.
+home: https://github.com/giantswarm/releases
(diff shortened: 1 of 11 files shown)


helm/releases-vmware/Chart.yaml
```diff
@@ -0,0 +1,6 @@
+apiVersion: v1
```
MarcelMue, line 1: I think you will need more changes to actually make this "deployable"

MarcelMue: See also here:
https://github.com/giantswarm/releases/blob/760c7d726591e84ded633847c7d6907d0cfbe077/.circleci/config.yml#L29

QuentinBisson: it is highly likely but I cannot really test is anyway :D

QuentinBisson: It should be fixed now :D

tfussell:
Closing in favor of https://github.com/giantswarm/releases/pull/804

PR #645, opened 2021-03-15
Use workload cluster naming not tenant in docs

rossf7:
Noticed we're still using tenant cluster in a couple of places in the README and PR template.


stone-z (approved):
:+1:

PR #661, opened 2021-03-31
Release1410

paurosello:


alex-dabija (approved):
The PR is fine and it can be merged after the app-operator is updated because the validation step is failing.

PR #665, opened 2021-04-06
require new external-dns

whites11:
require externaldns 2.3.0 for all new azure and AWS releases

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -13,6 +13,10 @@ releases:
         - name: coredns
           version: ">= 1.4.0"
           issue: https://github.com/giantswarm/giantswarm/issues/13771
+    - name: "> 13.1.0 > 14.0.0 > 14.1.0"
+      requests:
+        - name: external-dns
+          version: ">= 2.3.0"
     - name: "> 14.0.0"
       requests:
         - name: app-operator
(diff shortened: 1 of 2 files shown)


azure/requests.yaml
```diff
@@ -61,23 +63,6 @@ releases:
         - name: cluster-operator
           version: ">= 0.23.20"
           issue: https://github.com/giantswarm/roadmap/issues/187
-
-    - name: ">= 12.0.0 < 13.0.0"
```
tuommaki, line 65: Are these cleanup for non-existing releases? In the future might make sense to do these in separate PR :)

tuommaki (approved):

fiunchinho (approved):

azure/requests.yaml
whites11: yeah right, thanks

aws/requests.yaml
```diff
@@ -13,6 +13,10 @@ releases:
         - name: coredns
           version: ">= 1.4.0"
           issue: https://github.com/giantswarm/giantswarm/issues/13771
+    - name: "> 13.1.0 > 14.0.0 > 14.1.0"
```
njuettner, line 16: Shall we just name it `> 13.1.0` ?
```suggestion
    - name: "> 13.1.0"
```

whites11: My intent was to avoid it involving existing 14.0.x and 14.1.x releases but I probably didn't fulfil it anyway

njuettner: Ah I see 👍🏻 , all good

PR #692, opened 2021-04-20
Update the request to request net-export v1.10.0+ on R14+

ericgraf:
This PR is to change the request parameter to request the next release increase the net-exporter version.

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -1,11 +1,9 @@
 releases:
-    - name: ">= 14.1.1"
+    - name: ">= 15.0.0"
       requests:
         - name: net-exporter
           version: ">= 1.10.0"
           issue: https://github.com/giantswarm/giantswarm/issues/16574
-    - name: ">= 15.0.0"
-      requests:
         - name: app-operator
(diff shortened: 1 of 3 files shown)


aws/requests.yaml
```diff
@@ -25,6 +25,9 @@ releases:
         - name: chart-operator
           version: ">= 2.12.0"
           issue: https://github.com/giantswarm/giantswarm/issues/15213
+        - name: net-exporter
```
MarcelMue, line 28: I think we should put this into it's own bracket which is higher than the currently latest version.

ericgraf: Changed let me know if this is better.

aws/requests.yaml
```diff
@@ -1,4 +1,9 @@
 releases:
+    - name: ">= 14.1.1"
```
MarcelMue, line 2: I think putting it in the `>= 15.0.0` bracket makes more sense -in general the `validate` CI job should succeed.

ericgraf: changed :)

MarcelMue (approved):

PR #695, opened 2021-04-22
Request flatcar 2765+ in next releases

stone-z:
Towards https://github.com/giantswarm/giantswarm/issues/16533 to keep the updates rolling

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -16,6 +16,11 @@ releases:
         - name: coredns
           version: ">= 1.4.0"
           issue: https://github.com/giantswarm/giantswarm/issues/13771
+    - name: "> 14.1.1"
+      requests:
+        - name: containerlinux
+          version: ">= 2765.2.2"
+          issue: https://github.com/giantswarm/giantswarm/issues/14019
     - name: "> 13.1.0 > 14.0.0 > 14.1.0"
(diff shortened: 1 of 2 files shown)


aws/requests.yaml
```diff
@@ -16,6 +16,11 @@ releases:
         - name: coredns
           version: ">= 1.4.0"
           issue: https://github.com/giantswarm/giantswarm/issues/13771
+    - name: "> 14.1.1"
```
MarcelMue, line 19: WDYT about including it in 15.0.0 instead?

stone-z: For AWS and Azure this is actually only a bump of 2 release versions, so not as big a jump as it looks. I'm also fine with 15, just trying to keep new stuff as fresh as possible and cut down the tail a little bit

MarcelMue: I am just concerned if we will get pinged if something goes wrong and teams are only trying to release patches as everyone is focussed on CAPI.

MarcelMue (approved):
I am fine either way - both ways have pros and cons

aws/requests.yaml
stone-z: They can always add an exception if it blows up, no harm done

PR #707, opened 2021-05-05
request cert-manager 2.7.0 for next aws releases

ubergesundheit:
Next releases AWS releases should include cert-manager 2.7.0

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -16,6 +16,11 @@ releases:
         - name: coredns
           version: ">= 1.4.0"
           issue: https://github.com/giantswarm/giantswarm/issues/13771
+    - name: "> 14.1.1 > 13.1.1"
+      requests:
+        - name: cert-manager
+          version: ">= 2.7.0"
+          issue: https://github.com/giantswarm/giantswarm/issues/17029
     - name: "> 14.1.1"
       requests:
         - name: containerlinux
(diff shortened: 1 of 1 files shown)


aws/requests.yaml
```diff
@@ -16,6 +16,11 @@ releases:
         - name: coredns
           version: ">= 1.4.0"
           issue: https://github.com/giantswarm/giantswarm/issues/13771
+    - name: "> 14.1.1 > 13.1.1"
```
njuettner, line 19: Does the order of version matter?
`"> 14.1.1 > 13.1.1"` vs `"> 13.1.1 > 14.1.1"`

ubergesundheit: No idea. Who could know? For funsies lets reverse this and push again to see what CI thinks

ubergesundheit:
```suggestion
    - name: "> 13.1.1 > 14.1.1"
```

njuettner (approved):

aws/requests.yaml
njuettner: Not sure I only saw it some lines below the order was different 🙂 .

ubergesundheit: Its parsed by this function https://github.com/giantswarm/releases/blob/master/releases_test.go#L198 (called https://github.com/giantswarm/releases/blob/master/releases_test.go#L103)

ubergesundheit:
Thanks!

PR #709, opened 2021-05-07
Azure 15.0.0-beta1 release

whites11:

PR #710, opened 2021-05-07
Only limit app-operator for 14.X releases

MarcelMue:
Currently we limit `app-operator` to be `< 4.0.0` - I believe in the `15.0.0` releases we should include the newest app-operator.

@giantswarm/team-batman should know best

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -30,7 +30,7 @@ releases:
       requests:
         - name: external-dns
           version: ">= 2.3.0"
-    - name: "> 14.0.0"
+    - name: "> 14.0.0 < 15.0.0"
       requests:
         - name: app-operator
           version: ">= 3.2.1 < 4.0.0"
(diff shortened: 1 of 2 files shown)


aws/requests.yaml
```diff
@@ -30,7 +30,7 @@ releases:
       requests:
         - name: external-dns
           version: ">= 2.3.0"
-    - name: "> 14.0.0"
+    - name: "> 14.0.0 < 15.0.0"
```
tomahawk28, line 33: Which one is working? 
https://github.com/giantswarm/releases/pull/710/files#diff-8521c1dcb284f39dd6989ad25c0b64a724da61871c39dd085e985f055b01aaf9R7-R9
For better clarity, I think your changes make sense.

tomahawk28 (approved):

aws/requests.yaml
MarcelMue: One is for aws the other for azure - I have tried both with the latest app-operator now.

See the PR provided by Christian for Azure for example - https://github.com/giantswarm/releases/pull/709

stone-z (approved):
I don't have much context for app-operator but limiting the request this way is better practice IMO

PR #718, opened 2021-05-14
Azure 15.0.0-beta2 

whites11:
same as beta1 with downgraded flatcar


README.md
```diff
@@ -128,6 +128,7 @@ to all Giant Swarm installations.
 
 - v15
   - v15.0
+    - [v15.0.0-beta2]((https://github.com/giantswarm/releases/tree/master/azure/v15.0.0-beta2)
```
fiunchinho, line 131: There is an extra `(` in both betas.

whites11: good catch

fiunchinho (approved):
should we archive `v15.0.0-beta1` instead of just deprecate?

T-Kukawka (approved):

whites11:
> should we archive `v15.0.0-beta1` instead of just deprecate?

no, it's still in use.
After I release beta2, I'll upgrade `5jka7` and then archive

PR #728, opened 2021-05-25
Request external-dns v2.3.1 for AWS

rossf7:
Latest release from @webwurst for https://github.com/giantswarm/giantswarm/issues/15851 increases the memory limit from 50Mi to 100Mi.

I needed to apply this manually in a China cluster earlier. So just making sure it gets in the next releases.


rossf7:
Thx!

PR #731, opened 2021-05-28
Archive AWS release v9.0.9

alex-dabija:
It's unused and obsolete.

I identified the release as being unused by using cli.py script from [po-utils](https://github.com/giantswarm/po-utils). I also looked at the following graph: https://giantswarm.grafana.net/d/6Jt4OwqZk/clusters?viewPanel=39&orgId=1 .


anvddriesch (approved):
Nothing better than getting rid of 9s

PR #732, opened 2021-05-28
AWS 14.2.1

njuettner:


aws/v14.2.1/release.diff
```diff
@@ -0,0 +1,69 @@
...
+    name: metrics-server                                               name: metrics-server
+    version: 1.3.0                                                     version: 1.3.0
+  - name: net-exporter                                               - name: net-exporter
+    version: 1.10.1                                                    version: 1.10.1
+  - componentVersion: 1.0.1                                          - componentVersion: 1.0.1
+    name: node-exporter                                                name: node-exporter
+    version: 1.7.2                                                     version: 1.7.2
+  components:                                                        components:
+  - name: app-operator                                               - name: app-operator
+    releaseOperatorDeploy: false                                <
```
njuettner, line 43: @rossf7 can you have a look if this is correct? I just only made a copy via devctl from former release `14.2.0`

rossf7: @njuettner Yes this is correct. With cluster-operator 4.7.0 we create an app-operator per WC and its no longer managed by release-operator.

https://github.com/giantswarm/cluster-operator/blob/master/CHANGELOG.md#370---2021-03-15

njuettner:
I added a small change in tests for uniqueness. The problem is tests compare previous release and checks if anything has changed. It only checks version, however we only change the reference. This is a quick hack to also consider if previous release is deprecated, it will not compare it.

njuettner:
Ok tests are passing now 🎉

njuettner:
After fixing route53-manager in China it also works there 🎉 .
Good to go to release it.

anvddriesch (approved):

releases_test.go
```diff
@@ -164,6 +164,9 @@ func findReleases(provider string, archived bool) ([]v1alpha1.Release, error) {
 func releasesToIndex(releases []v1alpha1.Release) []versionbundle.IndexRelease {
 	var indexReleases []versionbundle.IndexRelease
 	for _, release := range releases {
+		if release.Spec.State != "active" {
```
alex-dabija, line 167: Is this change still required? We have a newer aws-operator in the release.

njuettner: Reverted 👍🏻

alex-dabija (approved):

PR #745, opened 2021-06-25
Update AWS CAPA release with latest Kyverno

MarcelMue:


tfussell:
Why is kyverno in the release instead of in aws-app-collection?

MarcelMue:
> Why is kyverno in the release instead of in aws-app-collection?

We do not want it on prod installations for AWS by default. So the release is set to deprecated -> it only gets deployed when we test CAPI

njuettner (approved):

anvddriesch (approved):

anvddriesch:
The policies would need to be updated individually if that is needed.

PR #763, opened 2021-07-15
use custom kiam and bump kyverno policies

calvix:
use custo kiam-app for v20.0.0 CAPA release from https://github.com/giantswarm/kiam-app/pull/84

use new kyverno policies which will add label to worker role to work with kiam-app


calvix:
tested, with this CAPA clusters will run all the apps without issues

anvddriesch (approved):
let's go!

PR #802, opened 2021-09-07
Make AWS v15.2.0 release active

alex-dabija:
Make AWS v15.2.0 release active in order to enable upgrades via Happa.


alex-dabija:
This is no longer required.

PR #809, opened 2021-09-13
Only push releases for changed providers

tfussell:
For https://github.com/giantswarm/giantswarm/issues/16770

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -1,8 +1,10 @@
 
 version: 2.1
 
+setup: true
+
 orbs:
-  architect: giantswarm/architect@4.3.0
+  path-filtering: circleci/path-filtering@0.0.3
 
 jobs:
   build:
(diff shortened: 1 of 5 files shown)


tfussell:
Tested changing one provider and all three providers. You can see the logs here https://app.circleci.com/pipelines/github/giantswarm/releases?branch=dynamic-config.

There are two possible issues:
- We can't use `requires` between separate workflows so PRs that change all 3 providers such as adding requests can potentially interfere during on parallel `git pull` and `git push`.
- We can't require status checks to pass because we don't know which one will run.

I think this is still an overall improvement for reducing build time and adding fewer unused charts to the `releases-test-catalog`.

MarcelMue (approved):
That's neat!

fiunchinho:
This is great! Thanks Thomas

tfussell:
I'm afraid this broke the rendered helm charts even though I was sure I tested it. I will retry later today or tomorrow and get someone else to verify with me.

PR #812, opened 2021-09-14
Only push releases for changed providers - fixed

tfussell:
Retrying reverted PR https://github.com/giantswarm/releases/pull/809

In that PR, I ensured that the helm charts were being pushed to `releases-test-catalog` as expected but not the content of the charts. `persist_to_workspace` only saves files in a single workflow. Because the new approach has one workflow per provider, the files from the `build` step were lost so the rendered helm charts were empty.

The new approach here is to move the `build` step into a reusable job, `render`, which can be called for each provider workflow with a parameter, `provider`. I have ensured that the resulting helm charts have the correct releases by changing one version in one release for each provider (in cce410bd941c8e3e37f1c539369c21dd59c09603 and 33e1cdd27396877ebcddebd4af3184bd93030860) and then checking the resulting helm charts in `releases-test-catalog` for the correct Release CRs contaning those changes:
- https://github.com/giantswarm/releases-test-catalog/blob/a613d05f337fb543d37e509e992253e7cfaae388/releases-aws-0.0.0-257f9474fe10b34af0c848c0a6bc4c06ce0073ab.tgz
- https://github.com/giantswarm/releases-test-catalog/blob/41b62db1cb66834a53ae5d86a8fca91a41bcd59f/releases-kvm-0.0.0-257f9474fe10b34af0c848c0a6bc4c06ce0073ab.tgz
- https://github.com/giantswarm/releases-test-catalog/blob/ef9946f804e448c76d75754cd54fe0a44f9efb97/releases-azure-0.0.0-257f9474fe10b34af0c848c0a6bc4c06ce0073ab.tgz

The CircleCI workflows can be viewed here https://app.circleci.com/pipelines/github/giantswarm/releases?branch=dynamic-config-fixed


whites11 (approved):
Seems ok

rossf7 (approved):
Looks good and nice catch on `persist_to_workspace` 👍

PR #825, opened 2021-10-01
AWS 15.2.2

njuettner:

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -1,5 +1,5 @@
 releases:
-    - name: "> 15.2.1"
+    - name: "> 15.2.2"
       requests:
         - name: external-dns
           version: ">= 2.5.0"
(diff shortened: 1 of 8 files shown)


aws/requests.yaml
```diff
@@ -1,5 +1,5 @@
 releases:
-    - name: "> 15.2.1"
+    - name: "> 15.2.2"
```
AverageMarcus, line 2: Is this right? Aren't those dependencies still valid for 15.2.1? (or have I misunderstood)

njuettner: The reason why I bumped it was simply because we had to update all the components, currently cert-manager is broken. In order to release it quickly without including all the components I simply bumped the version.

njuettner: inside the requests.yaml we add components which needed to be higher than a certain version if you would create a new release.

alex-dabija (approved):

PR #831, opened 2021-10-06
renamed capa-iam-controller to capa-iam-operator

AverageMarcus:
Related issue: https://github.com/giantswarm/roadmap/issues/468

Blocked on release of https://github.com/giantswarm/capa-iam-controller/pull/80


njuettner (approved):

aws/v20.0.0/release.yaml
```diff
@@ -49,10 +49,10 @@ spec:
     releaseOperatorDeploy: true
     version: 0.1.1
   - catalog: control-plane-catalog
-    name: capa-iam-controller
-    reference: 0.2.0
+    name: capa-iam-operator
+    reference: 0.3.0
```
njuettner, line 53: I think we do not need the reference anymore

AverageMarcus: I wasn't sure what `reference` was used for so I just followed what was already there. If it's safe to remove I can do that.

njuettner: it's also fine leaving it right now, no biggie :)

PR #859, opened 2021-11-15
AWS 16.1.0

njuettner:
Addresses: https://github.com/giantswarm/roadmap/issues/526


ubergesundheit:
Can we downgrade cert-exporter to the previous version `1.8.0` ? `2.0.0` is causing problems when upgrading.

alex-dabija (approved):

PR #861, opened 2021-11-25
Update capi/capa components

njuettner:
This will fix running multiple apps via argo and release. 

There was a bug which stopped helm installing the app for CRD webhook and
dynamic webhooks plus controller. It was only able to install CRD
webhook via argo but failed deploying the release apps.

We also had to rename the tags for the CRD webhook (ending *-crd)
because the management webhook doesn't allow to have the same tag for
the same app twice.

This madness will hopefully end when using v1alpha4 and above. 

Issue: https://github.com/giantswarm/giantswarm/issues/19415


whites11 (approved):
gr8

PR #864, opened 2021-11-29
Request cert-operator v1.1.0 for Azure

rossf7:
Towards https://github.com/giantswarm/giantswarm/issues/19913

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,4 +1,8 @@
 releases:
+    - name: "> 16.0.2"
+      requests:
+        - name: cert-operator
+          version: ">= 1.1.0 < 1.2.0"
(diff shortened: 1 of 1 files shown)


azure/requests.yaml
```diff
@@ -1,4 +1,8 @@
 releases:
+    - name: "> 16.0.2"
+      requests:
+        - name: cert-operator
+          version: ">= 1.1.0 < 1.2.0"
```
rossf7, line 5: cert-operator 1.1.0 uses the config repo rather than draughtsman values.

I tested with 1.2.0 but cluster creation failed due to this check

> Add check to ensure that the Cluster resource is in the same namespace as the certConfig before creating the secret there.

fiunchinho (approved):

PR #865, opened 2021-11-29
Request cert-operator v1.1.0 for KVM

rossf7:
Towards giantswarm/giantswarm#19913

### DIFF
--- a/kvm/requests.yaml
+++ b/kvm/requests.yaml
@@ -3,3 +3,5 @@ releases:
       requests:
         - name: cert-exporter
           version: ">= 2.0.0"
+        - name: cert-operator
+          version: ">= 1.1.0 < 1.2.0"


kvm/requests.yaml
```diff
@@ -3,3 +3,5 @@ releases:
       requests:
         - name: cert-exporter
           version: ">= 2.0.0"
+        - name: cert-operator
+          version: ">= 1.1.0 < 1.2.0"
```
rossf7, line 7: cert-operator 1.1.0 uses the config repo rather than draughtsman values.

I tested with 1.2.0 on Azure but cluster creation failed due to this check

> Add check to ensure that the Cluster resource is in the same namespace as the certConfig before creating the secret there.

yulianedyalkova (approved):

PR #868, opened 2021-11-30
Request Flatcar 2983.2.1 in AWS 16.1.x or above

stone-z:
Please bump Flatcar to include AWS SSM Agent support


calvix (approved):
that will be fun :p

PR #906, opened 2022-01-17
Request external-dns 2.8.0 in upcoming aws and azure platform releases

ubergesundheit:
Please include external-dns 2.8.0 in upcoming AWS and Azure platform releases.

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -9,6 +9,10 @@ releases:
       requests:
         - name: containerlinux
           version: "<= 2905.2.6"
+    - name: "> 16.3.0 > 16.1.0 > 16.0.1"
+      requests:
+        - name: external-dns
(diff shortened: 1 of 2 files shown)


aws/requests.yaml
```diff
@@ -9,6 +9,10 @@ releases:
       requests:
         - name: containerlinux
           version: "<= 2905.2.6"
+    - name: "> 16.3.0 > 16.1.0 > 16.0.1"
```
njuettner, line 12: Can we change it to 16.3.1? We have a small fix for EC2 instances which shouldn't have any other changes? Pretty please?
```suggestion
    - name: "> 16.3.1 > 16.1.0 > 16.0.1"
```

ubergesundheit: sure :)

njuettner (approved):

PR #911, opened 2022-01-28
Delete OpenStack and vSphere releases

tfussell:
As of cluster-apps-operator v1.2.0, releases are no longer used anywhere in the OpenStack platform. Removing vSphere as well as it will use the same structure as OpenStack.


pipo02mix (approved):
makes sense

rossf7 (approved):
LGTM

MarcelMue (approved):

PR #925, opened 2022-02-03
AWS v17.0.0

njuettner:
AWS v17.0.0


aws/v17.0.0/announcement.md
```diff
@@ -0,0 +1,11 @@
+**Workload cluster release v17.0.0 for AWS is available**. It provides support for Kubernetes 1.22, has Control Groups v2 enabled by default and includes the Vertical Pod autoscaler. Further details can be found in the [release notes](https://docs.giantswarm.io/changes/workload-cluster-releases-aws/releases/aws-v17.0.0/).
+
+> **_Warning:_** Kubernetes v1.22 removed certain APIs and features. More details are available in the [upstream blog post](https://kubernetes.io/blog/2021/07/14/upcoming-changes-in-kubernetes-1-22/).
+
+> **_Warning:_** `rpcbind` is disabled by default to mitigate security risks. NFS `v2` and `v3` are not supported anymore. Please double check if any of your application still use any of those versions before you upgrade.
```
whites11, line 5: rpcbind's most common use case is NFS v2 and v3 but it could have been used for other stuff in theory. Is it worth mentioning? as it is so far it seems only NFS is affected which is not true

alex-dabija: Good point! I updated it with the following wording:
```
> **_Warning:_** `rpcbind` is disabled by default to mitigate security risks. Any application which requires it will no longer work. NFS `v2` and `v3` are such applications and are no longer supported. Please, check if any you have any application which depend on `rpcbind` before you upgrade.
```

What do you think?

whites11 (approved):

PR #942, opened 2022-02-21
Bump core-dns version in requests

AverageMarcus:
Signed-off-by: Marcus Noble <github@marcusnoble.co.uk>

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -1,4 +1,8 @@
 releases:
+    - name: "> 17.0.0"
+      requests:
+        - name: core-dns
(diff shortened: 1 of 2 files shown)


aws/requests.yaml
```diff
@@ -1,4 +1,8 @@
 releases:
+    - name: "> 17.0.0"
```
whites11, line 2: worth doing the same for > 16.1.1 as well?

AverageMarcus: If you think so. I'm not sure how previous versions are handled.

whites11 (approved):

PR #966, opened 2022-04-20
Requests: Fix indentation.

Gacko:
See https://github.com/giantswarm/fmt/blob/master/helm/formatting_guidelines.md#indentation.


kubasobon (approved):
LGTMIIWFY

PR #968, opened 2022-04-20
Requests: Fix patterns.

Gacko:
It seems those release patterns got a bit misunderstood in the past. I fixed them and attached more details and code snippets in the corresponding commit messages.

~~To make sure we're not breaking tests for already released versions, I fixed the pattern to what they are actually doing instead of what they were intended to do.~~

I ported back `kube-state-metrics` and `cert-manager` as requested by @ubergesundheit.

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -15,6 +15,12 @@ releases:
     version: ">= 2.9.0"
   - name: chart-operator
     version: ">= 2.20.1"
+- name: "> 17.1.0 < 17.2.0"
+  requests:
+  - name: kube-state-metrics
+    version: ">= 1.9.0"
+  - name: cert-manager
+    version: ">= 2.13.0"
 - name: "> 17.0.0"
   requests:
   - name: coredns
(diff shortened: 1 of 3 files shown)


ubergesundheit:
I would suggest pinging the right people (those who made the original requests) here.

Gacko:
So... one of them is you. The other one was @whites11, IIRC.

aws/requests.yaml
```diff
@@ -1,5 +1,5 @@
 releases:
-- name: "> 16.1.1 > 16.3.1 > 16.4.1 > 17.1.0 > 17.2.0"
+- name: "> 17.2.0"
```
ubergesundheit, line 2: Will this make sure kube-state metrics 1.9.0 or higher will end up in lets say release 16.4.2 ?

Gacko: No it doesn't. As mentioned in the PR description I only fixed the patterns to what they are actually doing, not to what they were intended to do. So there already may be some 16.x releases which are not applying to this request. Implementing what this pattern was intended to do could break already released versions.

AverageMarcus: No, you'll need to use `||` to define multiple ranges

whites11 (approved):

Gacko:
@ubergesundheit asked me to port kube-state-metrics and cert-manager back to new minor or patch releases of older major releases. So maybe you take another look at this latest commit.

aws/requests.yaml
Gacko: I added another commit which ensures what you requested :)

ubergesundheit (approved):
Looks good, I would appreciate if you could add/update the documentation either in this repository or https://intranet.giantswarm.io/docs/product/releases/requesting-changes-in-next-platform-release/

PR #957, opened 2022-03-24
add azure release 17.1.0

whites11:
@T-Kukawka you can use this as base for 17.1.0.
You only need to bump azure-operator to whatever version Nikola will come up with.
We also need to ensure azure-operator has k8scloudconfig >= 13.1.0

PR #991, opened 2022-06-13
Prepare azure release

uvegla:
The `chart-operator` version of at least [v2.21.1](https://github.com/giantswarm/chart-operator/releases/tag/v2.21.1) is needed for the `NOTES` column feature of `kubectl gs` to dispaly meaningful informationt.

There are related changes in `app-operator` as well and some nice fixes like the `VPA` update or the Split Helm Client for security purposes.

Bumped `cluster-operator` as well to release fix for handling App Bundles correctly in `v4.3.0`.

Extra: Requested bumps of other apps and operators.

### DIFF
--- a/README.md
+++ b/README.md
@@ -182,6 +182,7 @@ to all Giant Swarm installations.
 
 - v17
   - v17.0
+    - [v17.0.2](https://github.com/giantswarm/releases/tree/master/azure/v17.0.2)
     - [v17.0.1](https://github.com/giantswarm/releases/tree/master/azure/v17.0.1)
     - [v17.0.0](https://github.com/giantswarm/releases/tree/master/azure/v17.0.0)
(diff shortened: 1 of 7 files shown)


whites11:
hey @uvegla we are already working on a new azure release that bumps all components.
We'd rather deliver just that release rather than another patch release in 17.0.

See https://github.com/giantswarm/releases/pull/957 and check if you have the desired versions there

whites11 (changes requested):
We'd love to avoid this release and release 17.1.0 instead

uvegla:
> hey @uvegla we are already working on a new azure release that bumps all components. We'd rather deliver just that release rather than another patch release in 17.0.
> 
> See #957 and check if you have the desired versions there

Awesome, thank you! Closing this and moving conversation there. 👍🏻

PR #996, opened 2022-06-23
Fix azure 17.0.1 date

T-Kukawka:

### DIFF
--- a/azure/v16.1.3/release.yaml
+++ b/azure/v16.1.3/release.yaml
@@ -52,7 +52,7 @@ spec:
     version: 3.15.5
   - name: etcd
     version: 3.4.18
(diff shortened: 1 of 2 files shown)


azure/v16.1.3/release.yaml
```diff
@@ -52,7 +52,7 @@ spec:
     version: 3.15.5
   - name: etcd
     version: 3.4.18
-  date: "2022-05-19T11:27:23Z"
+  date: "2022-06-06T11:27:23Z"
```
whites11, line 55: why?

T-Kukawka: because it was released that date not on  19th  and in docs it shows 19th

whites11 (approved):

PR #998, opened 2022-06-28
Upgrade kube-state-metrics to v1.11.0

TheoBrigitte:


whites11 (changes requested):
Nope, the files you want to change are https://github.com/giantswarm/releases/blob/master/aws/requests.yaml and https://github.com/giantswarm/releases/blob/master/azure/requests.yaml

TheoBrigitte:
@whites11 is this better ?

azure/requests.yaml
```diff
@@ -3,6 +3,8 @@ releases:
   requests:
   - name: external-dns
     version: ">= 2.14.0"
+  - name: kube-state-metrics
```
whites11, line 6: 17.1.0 is already out, so by changing this block you are actually requiring a change in an existing release. 
That is why the CI is failing.

You need a new block for `> 17.1.0`

kvm/requests.yaml
```diff
@@ -4,7 +4,7 @@ releases:
   - name: chart-operator
     version: ">= 2.20.1"
   - name: kube-state-metrics
-    version: ">= 1.9.0"
+    version: ">= 1.11.0"
```
whites11, line 7: probably same as azure for kvm

TheoBrigitte: but what's the block condition then? we already have `> 16.2.0`

whites11: yeah correct, this one is ok

whites11 (approved):
Perfect, thank you!

PR #1007, opened 2022-07-11
AWS v18.0.0-alpha1

njuettner:
Release Issue: https://github.com/giantswarm/roadmap/issues/846
Roadmap Issue for Kubernetes 1.23: https://github.com/giantswarm/roadmap/issues/1137


uvegla:
@njuettner I bumped `app-operator` to `v6.3.0`.

whites11:
e2e tests green, safe to be releases from my POV

aws/v18.0.0-alpha1/README.md
```diff
@@ -0,0 +1,774 @@
+# :zap: Giant Swarm Release v18.0.0-alpha1 for AWS :zap:
+
+This release provides support for Kubernetes 1.23 and replaces the use of AWS CNI with Cilium. Release also upgrades most of running components which can be found further in release notes.
+
+> **_Warning:_** This is an **`alpha preview release`** intended only for testing Kubernetes v1.23 changes and replacement of AWS CNI with Cilium. Upgrading to or from this version is not supported.
```
whites11, line 5: We need to point out a few notes about Cilium I think.
The main change to me is how NetworkPolicies work. More specifically network policies that reference pod or service cidrs won't work any more see https://docs.cilium.io/en/stable/concepts/kubernetes/policy/#networkpolicy-state

alex-dabija:
> `Pod Subnet` which can be used to create ENIs for peering with VPCs is removed with Cilium. Please pay attention if any of the workloads depend on that connection as it will have to be adjusted accordingly.

There are multiple such subnets, one for each availability zone. I would rephrase it something like:

> The AWS CNI pod subnets are no longer used by Cilium and are removed. Please, add custom routes with the node subnet(s) CIDR(s) instead of the AWS CNI pod subnets CIDR before upgrading to this release.

alex-dabija:
@whites11, @paurosello or @silenteh Could you confirm if what I just suggested in my previous comment makes sense? Thanks!

whites11:
> @whites11, @paurosello or @silenteh Could you confirm if what I just suggested in my previous comment makes sense? Thanks!

The subnets are NOT removed in this release. They are still there, created but unused

PR #1011, opened 2022-07-27
AWS: Request cert-manager >= 2.15.2.

Gacko:
Fixes an issue introduced in v2.15.1 and occurring during cluster upgrades.

Fresh installs of AWS v17.4.1 work, upgrades from any version below to v17.4.1 may break. So we probably should release v17.4.2 soon so customers upgrading their workload clusters don't run into this issue.

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -7,6 +7,8 @@ releases:
     version: ">= 4.3.0"
   - name: chart-operator
     version: ">= 2.25.0"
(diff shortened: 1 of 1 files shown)


whites11 (approved):

whites11:
@Gacko do we need the same for Azure or is this an AWS-only change?

Gacko:
`cert-manager` is default on AWS only.

PR #1038, opened 2022-09-09
Request `kube-state-metrics` >= 1.13.0.

Gacko:


QuentinBisson (approved):

QuentinBisson:
Can you do the same in all default apps also ? e.g openstack-default-apps ?

Gacko:
Was already about to do so but not sure where, yet. Your question just pointed me to the right repos.

PR #1049, opened 2022-10-11
add azure release 18.0.2

whites11:
wip release for internal use only

### DIFF
--- a/README.md
+++ b/README.md
@@ -196,6 +196,7 @@ to all Giant Swarm installations.
 
 - v18
   - v18.0
+    - [v18.0.2](https://github.com/giantswarm/releases/tree/master/azure/v18.0.2)
     - [v18.0.1](https://github.com/giantswarm/releases/tree/master/azure/v18.0.1)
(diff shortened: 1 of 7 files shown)


Gacko (approved):

Gacko:
@whites11 @T-Kukawka Mind giving additional approval? I tested it with several clusters yesterday and it's a huge improvement even though I still spotted a single cluster not coming up. Calico CRD installer runs at the very beginning and even other resources like charts come up much faster now.

T-Kukawka:
Yes, first i want to create release issue etc

Gacko:
I basically don't care, that's up to you guys to decide.

Customers have no advantage on existing clusters and on new clusters it is not a show stopper - just annoying when it takes several attempts until the cluster is up.

I would do it just to have a consistent state on all installations and no "temporary" releases on test installations.

T-Kukawka:
We have decided not to release this part for customers. Release is available internally. There is little advantage of upgrading to that release especially  that the nodes will be rolled for 'no reason' in existing clusters. We suspect no new cluster from customers coming up any time soon. Closing.

PR #1050, opened 2022-10-12
Azure: Request `azure-operator` v6.0.3.

Gacko:
This is mostly a suggestion since we decided not to [release it with Azure v18.0.2](https://github.com/giantswarm/releases/pull/1049), yet. So declining this PR is also fine. :)


whites11:
Thanks @Gacko but this won't be necessary.
Next azure release will be v19 and it'll need a whole new version of azure-operator.

PR #1064, opened 2023-01-11
Update requests.yaml

mcbenjemaa:
Upgrade VPA,
In order to fix memory issues on various clusters.
Ref: https://github.com/giantswarm/giantswarm/issues/25180

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -11,6 +11,10 @@ releases:
     version: ">= 2.17.1"
   - name: observability-bundle
     version: ">= 0.1.4"
+- name: "> 18.1.0"
+  requests:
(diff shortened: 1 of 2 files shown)


QuentinBisson (approved):
Can you do azure also?

whites11 (changes requested):
you can't do this. You are changing the request for a release selector `> 17.3.3` that affects several releases already out there.
this means automation will break.
What you can do is make your request to the next release.

current latest is 18.1.0, so you need to add the requirement to `>18.1.0`

whites11 (approved):

whites11:
> shall i merge?

at your leisure

PR #1095, opened 2023-03-19
add-observability-bundle-requests

QuentinBisson:
This PR requests the use of the latest observability-bundle for aws releases >= 18.2.1 as it contain a useful fix for Atlas

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -21,6 +21,8 @@ releases:
   requests:
   - name: coredns
     version: ">= 1.14.2"
+  - name: observability-bundle
+    version: ">= 0.2.0"
 - name: "> 18.1.1"
   requests:
   - name: vertical-pod-autoscaler
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -9,8 +9,6 @@ releases:
     version: ">= 1.14.0"
   - name: cert-manager
     version: ">= 2.17.1"
-  - name: observability-bundle
-    version: ">= 0.2.0"
   - name: vertical-pod-autoscaler-app
     version: ">= 3.2.0"
   - name: vertical-pod-autoscaler-crd


aws/requests.yaml
```diff
@@ -21,6 +21,8 @@ releases:
   requests:
   - name: coredns
     version: ">= 1.14.2"
+  - name: observability-bundle
+    version: ">= 0.2.0"
```
hervenicol, line 25: Does `>=` mean future releases will always retrieve the latest version of o11y-bundle when released?

QuentinBisson: I think that's thé Idea yes but let's make sure we update that 😅

whites11: means if we have 0.2.0 CI will not break and we might NOT bump any further. If you need a newer version you might want to bump in the requests

QuentinBisson: Can I have an approval still?

aws/requests.yaml
```diff
@@ -21,6 +21,8 @@ releases:
   requests:
   - name: coredns
     version: ">= 1.14.2"
+  - name: observability-bundle
```
whites11, line 24: if you add this here, there is no need to keep having the same clause at line 12 and 13 (since > 18.2.1 also includes >= 19.0.0). Please remove either one

QuentinBisson: Well, when you created 18.2.1 you did not take it when it was set for 19.0.0 so I'm not sure what to do here

whites11: of course not.
if you set it to the `> 18.2.1` block then it's valid for 19.0.0 as well
if you set it to the `> 19.0.0` block, it's NOT valid for 18.2.1.

so you might want to remove it from the `> 19.0.0` block in order for your request to be valid for both 18.2.2+ and 19.0.0+

PR #1097, opened 2023-03-21
Requests to bump app-operator to v6.6.3

uvegla:
With `app-operator` version `v6.6.3` we adjusted resource usage and VPA: https://github.com/giantswarm/app-operator/releases/tag/v6.6.3

### Checklist
- [ ] Roadmap issue created

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -14,7 +14,7 @@ releases:
   - name: cluster-operator
     version: ">= 5.0.0"
   - name: app-operator
-    version: ">= 6.4.1"
+    version: ">= 6.6.3"
   - name: observability-bundle
     version: ">= 0.1.4"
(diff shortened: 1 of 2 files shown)


azure/requests.yaml
```diff
@@ -14,7 +14,7 @@ releases:
   - name: cluster-operator
     version: ">= 5.0.0"
   - name: app-operator
-    version: ">= 6.4.1"
+    version: ">= 6.6.3"
```
whites11, line 17: you can't do this. 19.0.1 is already releases, so you are requesting to bump an existing release, and this is not allowed. Please add this request in the >19.0.1 section above

uvegla: Thanks! Updated 👍🏻

whites11 (approved):

PR #1183, opened 2023-10-04
Request bump for `chart-operator` and adding `chart-operator-extensions` that depends on `prometheus-operator-crd`

uvegla:
Towards: https://github.com/giantswarm/giantswarm/issues/27558

### Checklist
- [ ] Roadmap issue created

### DIFF
--- a/aws/requests.yaml
+++ b/aws/requests.yaml
@@ -1,5 +1,5 @@
 releases:
-- name: "> 19.1.0"
+- name: "> 19.1.1"
   requests:
     - name: chart-operator
       version: ">= 3.0.0"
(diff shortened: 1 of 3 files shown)


aws/requests.yaml
```diff
@@ -1,4 +1,11 @@
 releases:
+- name: "> 19.1.0"
```
whites11, line 2: Unfortunately aws release 19.1.1 is already crafted and tested and we can't change it any more

```suggestion
- name: "> 19.1.1"
```

uvegla: All right, thanks!

PR #1218, opened 2024-03-07
Bump observability-bundle to 1.3.0

marieroque:
towards https://github.com/giantswarm/giantswarm/issues/29665

### Checklist
- [x] Roadmap issue created


T-Kukawka (approved):

aws/requests.yaml
```diff
@@ -2,7 +2,7 @@ releases:
 - name: ">= 20.0.0"
   requests:
   - name: observability-bundle
-    version: ">= 1.2.3"
+    version: ">= 1.3.0"
```
whites11, line 5: we can't do this. If we do this, the being-released v20.0.0 will be invalid cc @T-Kukawka

whites11 (changes requested):
see my comment

aws/requests.yaml
nprokopic: What Chris said ☝️ 

We would need a new v20.1.0 for this change.

@marieroque how important and critical is the update to observability-bundle v1.3.0, more precisely, how important and critical is the update of kube-prometheus-stack from 9.0.0 to 9.1.0 that happened in observability-bundle v1.3.0? Do we need this in order to solve some issues, or is it a new feature release?

@T-Kukawka we should keep v20.x.x as slow as possible, almost frozen, with mostly adding bugfixes, right?

PR #1226, opened 2024-04-05
Patch v19.1.1 and v19.2 releases with security-bundle v0.19.3

fhielpos:
> [!CAUTION]
> DO NOT MERGE.
>
> I plan to merge this at a sensible time, when I can monitor the rollout on existing clusters.

Towards: https://github.com/giantswarm/giantswarm/issues/30472

We had a bug in exception-recommender that leaves some objects inside the cluster with finalizers. These finalizers can block namespace deletion.

This version patches exception-recommender with the fix.


fhielpos:
In case this is approved: I don't plan to merge until I'm around so I can monitor the rollout.

whites11 (approved):

PR #1227, opened 2024-04-05
Patch v19.3 and v20 releases with newest security-bundle v1.6.3

fhielpos:
> [!CAUTION]
> DO NOT MERGE.
>
> I plan to merge this at a sensible time, when I can monitor the rollout on existing clusters.
Towards: https://github.com/giantswarm/giantswarm/issues/30473

We did some improvements to the security-bundle, which some customers need without going through an extra v19.3.2 upgrade. This PR bumps the security-bundle versions from v19.3.1 to v20.

We already agreed with said customer that this change was going to happen.


fhielpos:
Side note: v20 doesn't need to happen yet. It's mainly to keep things aligned.

whites11 (approved):

PR #1263, opened 2024-06-04
Add Releases SDK

nprokopic:
Towards https://github.com/giantswarm/roadmap/issues/3466 and https://github.com/giantswarm/roadmap/issues/3473

### What is this?

This PR adds Releases SDK, which for now just contains a Release struct. The idea is to have a small Releases SDK go module which can then be used from:
- operators,
- admission controllers,
- test frameworks,
- CLI tools,
- wherever we need to use release from go code.

### Why do this here?

- While we want to use the releases repo for the CAPI releases, we don't want to use literally every part of the old process.
  - Release CRD is ATM defined in [releases-operator](https://github.com/giantswarm/release-operator/tree/master/api/v1alpha1). Adding this as a dependency would pull in the entire go dependency tree with way too many outdated vintage components, which would bring in dependency hell into a component that just needs a `Release` struct. IMO we need a new source of truth for the `Release` struct, and having that in a simple go module (and without the burden of outdated vintage go dependencies) is the way to go.
- We already have go tests here in releases repo.
- Having Releases SDK in the same repo as release manifests (already in the repo) and go unit tests (already in the repo) would accelerate development and testing a lot, as we could do multiple changes in the single PR and test everything together.

### Future work

The only dependency of the Release SDK is:
```
k8s.io/apimachinery `v0.26.15`
```

It is using the latest `v0.26.x` because the first use case for the Releases SDK would be test frameworks where we use `v0.26.x` ATM.

IMO we should be able to remove even this, and implement everything with go generics and work with interfaces, so that clients that use Releases SDK can use different versions of upstream k8s libraries.

### Checklist
- [x] Roadmap issue created


AverageMarcus (approved):

njuettner (approved):

nprokopic:
Added missing markdown files for the Releases SDK:
- README.md
- CHANGELOG.md
- CODEOWNERS.md

PR #1276, opened 2024-06-20
Add AWS (CAPA) v25.0.0

nprokopic:
This PR adds AWS (CAPA) v25.0.0 release.

App versions are the same as in previously-tested v25.0.0-alpha.3 and cluster-aws release v0.79.0 (the latest cluster-aws release that is not using releases).

Compared to v25.0.0-alpha.3, v25.0.0 only changes cluster-aws version to v1.0.0 (cluster-aws release PR https://github.com/giantswarm/cluster-aws/pull/666).

### DIFF
--- a/capa/v25.0.0/release.yaml
+++ b/capa/v25.0.0/release.yaml
@@ -1,0 +1,122 @@
+apiVersion: release.giantswarm.io/v1alpha1
+kind: Release
+metadata:
+  name: aws-25.0.0
+spec:
+  apps:
+  - name: aws-ebs-csi-driver
+    version: 2.30.1
+    dependsOn:
+    - cloud-provider-aws
+  - name: aws-ebs-csi-driver-servicemonitors
+    version: 0.1.0
+    dependsOn:
+    - cert-manager
+  - name: aws-pod-identity-webhook
+    version: 1.16.0
+    dependsOn:
+    - cert-manager
+  - name: capi-node-labeler
+    version: 0.5.0
+  - name: cert-exporter
+    version: 2.9.0
+    dependsOn:
+    - kyverno
+  - name: cert-manager
+    version: 3.7.6
+    dependsOn:
+    - prometheus-operator-crd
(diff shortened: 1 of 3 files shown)


capa/v25.0.0/release.yaml
```diff
@@ -0,0 +1,122 @@
+apiVersion: release.giantswarm.io/v1alpha1
+kind: Release
+metadata:
+  name: aws-25.0.0
```
njuettner, line 4: what is the naming convention of EKS cluster? will this be different? will it align and we verify somewhere else if it will be eks or capa?

njuettner: and what is the reason for not calling it capa-25.0.0?

njuettner (approved):
Not blocking here

capa/v25.0.0/release.yaml
paurosello: Not sure what is best, though I think the 2nd one is easier to read:

capa-v25.0.0
capa-v25.0.1
eks-v25.0.0
capz-v25.0.0
capz-v25.0.1
aks-v25.0.0

vs

aws-v25.0.0
aws-v25.0.1
eks-v25.0.0
azure-v25.0.0
azure-v25.0.1
aks-v25.0.0





[09:11](https://gigantic.slack.com/archives/D013546TA3D/p1718953873178509)

paurosello (approved):

capa/v25.0.0/release.yaml
nprokopic: IIUC long term `capa` will become `aws`, as soon as we completely migrate away from vintage https://github.com/giantswarm/roadmap/issues/1543.

Also long-term, even mid-term maybe, we should re-work Release CRD, so we can set provider name in some different way, like in spec, so CR name will not be that important like here, as here we are using Release CRD that has been designed for vintage, so we had to put the provider somewhere.

PR #1278, opened 2024-06-21
Fix draft tag action

nprokopic:
Fix GitHub action that is creating draft tags for WIP releases from PRs.

### DIFF
--- a/.github/workflows/create-draft-release.yaml
+++ b/.github/workflows/create-draft-release.yaml
@@ -38,12 +38,20 @@ jobs:
             git push --delete origin "$draft_tag_name"
             # Create new tag locally
             git tag -fa "$draft_tag_name" -m "AWS draft release $aws_release_version"
+            git push -f origin "$draft_tag_name"
           else
-            echo "Tag $draft_tag_name not found, will create it"
-            git tag -a "$draft_tag_name" -m "AWS (CAPA) draft release $aws_release_version"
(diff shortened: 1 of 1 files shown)


.github/workflows/create-draft-release.yaml
```diff
@@ -38,12 +38,20 @@ jobs:
...
             # Create new tag locally
             git tag -fa "$draft_tag_name" -m "AWS draft release $aws_release_version"
+            git push -f origin "$draft_tag_name"
           else
-            echo "Tag $draft_tag_name not found, will create it"
-            git tag -a "$draft_tag_name" -m "AWS (CAPA) draft release $aws_release_version"
+            echo "Tag $draft_tag_name not found"
+            
+            published_tag_name="aws/$aws_release_version"
+            published_release=$(gh release --repo giantswarm/releases view "$published_tag_name" --json "id,name,tagName,assets,isDraft,isPrerelease,url" 2>&1 || true)
```
njuettner, line 46: is `gh` included in github action?

njuettner (approved):

.github/workflows/create-draft-release.yaml
nprokopic: Yeah, I guess they add common tools there 

This worked before, it was just unnecessarily pushing draft tags when not needed. (maybe I just remove this, now with new functions to get releases from any git branch, but can't think about that decision today 🙈)

PR #1312, opened 2024-07-10
Custom renovate datasource

njuettner:
Towards: https://github.com/giantswarm/roadmap/issues/3547

Using release version in multiple repos via renovate wouldn't work the way how we use it right now because we need to distinguish the provider and also modifying the release version ([see example](https://github.com/giantswarm/workload-clusters-fleet/blob/main/management-clusters/gazelle/organizations/giantswarm-production/workload-clusters/operations.yaml#L17))

The idea here is to generate JSON via GH Action (Custom Renovate Datasource requires certain structure) for each provider so we can use it in each repo, example (not 100% accurate):

How the JSON looks like:

```
{
  "releases": [
    {
      "Version": "25.0.0",
      "IsDeprecated": false,
      "releaseTimestamp": "2024-06-20 18:00:00 +0000 UTC",
      "changelogUrl": "https://github.com/giantswarm/releases/blob/master/capa/v25.0.0/README.md",
      "isStable": true
    },
    {
      "Version": "26.0.0",
      "IsDeprecated": false,
      "releaseTimestamp": "2024-06-26 09:00:00 +0000 UTC",
      "changelogUrl": "https://github.com/giantswarm/releases/blob/master/capa/v26.0.0/README.md",
      "isStable": true
    }
  ],
  "sourceUrl": "https://github.com/giantswarm/releases",
  "changelogUrl": "https://github.com/giantswarm/releases/blob/main/README.md",
  "homepage": "https://giantswarm.io"
}
```

What's needed in repos to update `cluster_version`:
```
{
  "customDatasources": {
    "releases-capa": {
      "defaultRegistryUrlTemplate": "https://raw.githubusercontent.com/giantswarm/releases/master/capa/releases.json"
    }
  },
  "regexManagers": [
    {
      "fileMatch": ["something"],
      "matchStrings": ["cluster_release"],
      "datasourceTemplate": "custom.releases-capa",
      "depNameTemplate": "releases-capa",
      "extractVersionTemplate": "^(?<version>.*)$"
    }
  ]
}
```

The benefit is it can also be used by customer to update their repos when using Gitops.

The other idea was to use Github Tags as Datasource but you cannot rewrite the release version (e.g. capa-v26.0.0 -> 26.0.0).

### DIFF
--- a/.github/workflows/renovate-release-json.yaml
+++ b/.github/workflows/renovate-release-json.yaml
@@ -1,0 +1,63 @@
+name: Fetch Releases and Generate JSON
+
+on:
+  push:
+    branches:
+      - main
+  workflow_dispatch:
+
+jobs:
+  generate-json:
+    runs-on: ubuntu-latest
+    steps:
+      - name: Checkout repository
+        uses: actions/checkout@v3
+      
+      - name: "Setup Go Env"
+        uses: "actions/setup-go@v4"
+        with:
+          cache: true
+          go-version: "1.19.2"
+
+      - name: Generate JSON File
+        run: |
+          go run renovate.go
+
+      - name: "Ensure Private Key"
+        env:
+          BOT_SSH_KEY: '${{ secrets.BOT_SSH_KEY }}'
+        run: |
+          echo "$BOT_SSH_KEY" > /tmp/id_ed25519
+
+      - name: "Setup SSH Agent"
+        env:
+          SSH_AUTH_SOCK: "/tmp/ssh_agent.sock"
+        run: |
+          mkdir -p ~/.ssh
+          ssh-keyscan github.com >> ~/.ssh/known_hosts
+          ssh-agent -a ${SSH_AUTH_SOCK} > /dev/null
+          chmod 0600 /tmp/id_ed25519
+          ssh-add /tmp/id_ed25519
+
+      - name: "Setup Git Config"
+        run: |
+          git config user.name "${GITHUB_ACTOR}"
+          git config user.email "${GITHUB_ACTOR}@users.noreply.github.com"
+          git remote set-url origin git@github.com:${GITHUB_REPOSITORY}.git
+
+      - name: "Commit And Push"
+        env:
+          SSH_AUTH_SOCK: "/tmp/ssh_agent.sock"
+        run: |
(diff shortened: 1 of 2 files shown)


renovate.go
```diff
@@ -0,0 +1,99 @@
+package main
```
nprokopic, line 1: A bit late to the party. We could move this into some subdirectory, as releases repo structure is getting a bit messy (it has been already before this PR).

Nothing blocking in any case, as we can tidy it at any time.

njuettner: Good input, will do that in another PR 👍🏻

.github/workflows/renovate-release-json.yaml
```diff
@@ -0,0 +1,63 @@
...
+        env:
+          SSH_AUTH_SOCK: "/tmp/ssh_agent.sock"
+        run: |
+          mkdir -p ~/.ssh
+          ssh-keyscan github.com >> ~/.ssh/known_hosts
+          ssh-agent -a ${SSH_AUTH_SOCK} > /dev/null
+          chmod 0600 /tmp/id_ed25519
+          ssh-add /tmp/id_ed25519
+
+      - name: "Setup Git Config"
```
nprokopic, line 42: I think we have this simpler git config setup for GitHub actions https://github.com/giantswarm/releases/blob/cc714ffd0d84370138f68bc91a3dcb103afbbd34/.github/workflows/create-release.yaml#L51

.github/workflows/renovate-release-json.yaml
```diff
@@ -0,0 +1,63 @@
...
+        run: |
+          go run renovate.go
+
+      - name: "Ensure Private Key"
+        env:
+          BOT_SSH_KEY: '${{ secrets.BOT_SSH_KEY }}'
+        run: |
+          echo "$BOT_SSH_KEY" > /tmp/id_ed25519
+
+      - name: "Setup SSH Agent"
```
nprokopic, line 32: Why is this needed here?

.github/workflows/renovate-release-json.yaml
njuettner: In my opinion it should be committed by the owner, e.g. you merger CAPA Release v27.0.0. Releases.json should be updated from the one who merged it.

PR #1324, opened 2024-07-24
Add docs to Azure v25.0.0

yulianedyalkova:
### Checklist
- [ ] Roadmap issue created

### Triggering e2e tests

To trigger the E2E test for each new Release added in this PR add a comment with the following:

`/run releases-test-suites`

If you want to trigger conformance tests you can do so by adding a comment similar to the following:

`/run conformance-tests RELEASE_VERSION=v28.0.0 PROVIDER=capa`

For more details see the [README.md](/README.md#running-tests-against-prs)

### DIFF
--- a/azure/v25.0.0/README.md
+++ b/azure/v25.0.0/README.md
@@ -1,15 +1,15 @@
 # :zap: Giant Swarm Release v25.0.0 for Azure :zap:
 
-We are happy to announce our first release for Cluster API for Azure (CAPZ) that uses the new releases framework.
+We are happy to announce the first release for Azure that uses the new release framework.
 
 ## Migration to new releases flow
 
 In order to consume the new flow, the following two fields need to be manually adapted:
 
 * In ConfigMap `<cluster name>-userconfig` set `.Values.global.release` to the release version, e.g. `25.0.0`.
-* In App `<cluster name>` set the `version` to an empty string.
+* In App `<cluster name>` remove the `spec.version` field. In case of GitOps, Flux might complain that the app manifest is invalid as the `spec.version` field is mandatory. In that case, edit the live App CR and set `spec.version` to an empty string. That will unblock Flux and allow it reconcile successfully.
 
-And if you want to use `kubectl-gs` to create a cluster, you’d need to now specify the release version, e.g.:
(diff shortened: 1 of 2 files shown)


njuettner (approved):
lgtm

Gacko (approved):

Gacko:
Thank you! Someone needs to do it...

azure/v25.0.0/README.md
```diff
@@ -1 +1,16 @@
...
+We are happy to announce our first release for Cluster API for Azure (CAPZ) that uses the new releases framework.
+
+## Migration to new releases flow
+
+In order to consume the new flow, the following two fields need to be manually adapted:
+
+* In ConfigMap `<cluster name>-userconfig` set `.Values.global.release` to the release version, e.g. `25.0.0`.
+* In App `<cluster name>` set the `version` to an empty string.
+
+And if you want to use `kubectl-gs` to create a cluster, you’d need to now specify the release version, e.g.:
```
Gacko, line 12:
```suggestion
And if you want to use `kubectl-gs` to create a cluster, you'd need to now specify the release version, e.g.:
```

azure/v25.0.0/README.md
```diff
@@ -1 +1,16 @@
...
+# :zap: Giant Swarm Release v25.0.0 for Azure :zap:
+
+We are happy to announce our first release for Cluster API for Azure (CAPZ) that uses the new releases framework.
+
+## Migration to new releases flow
+
+In order to consume the new flow, the following two fields need to be manually adapted:
+
+* In ConfigMap `<cluster name>-userconfig` set `.Values.global.release` to the release version, e.g. `25.0.0`.
+* In App `<cluster name>` set the `version` to an empty string.
```
iuriaranda, line 10:
```suggestion
* In App `<cluster name>` remove the `spec.version` field.
* If using GitOps, after pushing the updated manifests, Flux might complain that the App manifest is invalid as the `spec.version` field is mandatory. If that's the case, edit the live App CR and set `spec.version` to an empty string. This will unblock Flux and let it reconcile successfully.
```

yulianedyalkova: @iuriaranda can we just default to setting the value to an empty string (how it was) rather than doing the whole dance of trying to remove it first etc? Or do we really want the field gone?

PR #1330, opened 2024-07-26
Fix GetLatestRelease

nprokopic:
This PR fixes the `GetLatestRelease` in the Releases SDK to correctly return the latest release according to semver, and not the latest published GitHub release. The function now searches for the latest release in the releases GitHub repo and not in GitHub releases.

After merging and releasing this, we should then bump the Releases SDK version that we use in e2e tests, in order to fix the currently broken CAPA tests.

### DIFF
--- a/sdk/builder.go
+++ b/sdk/builder.go
@@ -245,12 +245,14 @@ func (b *Builder) buildPreReleaseString() (string, error) {
 	}
 
 	// check if cluster app has a pre-release
-	clusterAppPreReleaseString, err := versionStringHasPreReleaseString(b.clusterApp.Version)
-	if err != nil {
-		return "", microerror.Mask(err)
-	}
-	if clusterAppPreReleaseString {
-		return preReleaseBuildFunc(), nil
+	if b.clusterApp != nil {
+		clusterAppPreReleaseString, err := versionStringHasPreReleaseString(b.clusterApp.Version)
+		if err != nil {
+			return "", microerror.Mask(err)
+		}
+		if clusterAppPreReleaseString {
+			return preReleaseBuildFunc(), nil
+		}
 	}
 
 	// check if apps have a pre-release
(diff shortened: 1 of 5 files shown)


sdk/builder.go
```diff
@@ -245,12 +245,14 @@ func (b *Builder) buildPreReleaseString() (string, error) {
 	}
 
 	// check if cluster app has a pre-release
-	clusterAppPreReleaseString, err := versionStringHasPreReleaseString(b.clusterApp.Version)
-	if err != nil {
-		return "", microerror.Mask(err)
-	}
-	if clusterAppPreReleaseString {
-		return preReleaseBuildFunc(), nil
+	if b.clusterApp != nil {
```
nprokopic, line 248: This is unrelated fix, as the nil pointer panic got uncovered here as I updated the unit tests.

sdk/builder_test.go
```diff
@@ -36,7 +36,7 @@ var _ = Describe("Builder", func() {
 		// set a cluster app override where the pre-release is different
 		const preReleasePrefix = "test"
 		releasesBuilder = releasesBuilder.
-			WithClusterApp("0.76.1-efdaba18e6e9866d62f3214f3d898b21c21e8b48", "").
+			WithClusterApp("1.0.0-efdaba18e6e9866d62f3214f3d898b21c21e8b48", "").
```
nprokopic, line 39: Updating the test here and in other places below, as now in unit tests the latest release is `v25.0.0` (and it used to be `v25.1.0-demo.0`), so cluster-aws version is different in the base release.

nprokopic:
I have updated the unit tests and unit tests work as expected now. Also I have tested this in a small go program to try to get really published latest releases for CAPA and CAPZ from the releases repo - and it works, now it is correctly returning v28.0.0 for CAPA and v25.0.0 for CAPZ.

Will merge and release this with a ping here, for the sake of fixing e2e tests as soon as possible tomorrow (when I am AFK, hence fixing this now).

PR #1355, opened 2024-08-29
CAPA v29.1.0: Define catalog for `cilium-crossplane-resources`.

Gacko:
This PR explicitly defines the catalog for the `cilium-crossplane-resources` app.

At the moment, `cluster-aws` is hard-coding this catalog anyways. So merging this PR will effectively not change any existing clusters running on this release.

But as I'm currently about to implement the same logic of looking up catalogs from the Release CR in the `cluster-aws` chart as we already have in the `cluster` chart, the Cluster Test Suites might fail as they are picking the latest available Release CR, patch the `cluster-aws` chart version to test and then execute their tests.

In this specific scenario `cilium-crossplane-resources` would be rendered with `catalog: default` as without this PR the latest Release CR does not explicitly define a catalog for it and therefore falls back to `default`.

**tl;dr:** This PR has no effect on production clusters, it is only required for development.

### Triggering E2E tests

`/run releases-test-suites`

`/run conformance-tests PROVIDER=capa RELEASE_VERSION=v29.0.0`

### DIFF
--- a/capa/v29.1.0/release.diff
+++ b/capa/v29.1.0/release.diff
@@ -33,6 +33,7 @@ spec:									spec:
   - name: cilium							  - name: cilium
     version: 0.25.1							    version: 0.25.1
   - name: cilium-crossplane-resources					  - name: cilium-crossplane-resources
+								>           catalog: cluster
     version: 0.1.0							    version: 0.1.0
   - name: cilium-servicemonitors					  - name: cilium-servicemonitors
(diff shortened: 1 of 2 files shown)


Gacko:
/run releases-test-suites TARGET_SUITES=./providers/capa/standard,./providers/capa/upgrade PREVIOUS_RELEASE=28.1.1 TARGET_RELEASES=aws-29.1.0

fiunchinho (approved):

PR #1419, opened 2024-09-24
CAPA: Release v29.2.0.

Gacko:
### Checklist

- [x] Roadmap issue created
- [x] Release uses latest stable Flatcar
- [x] Release uses latest Kubernetes patch version

### Triggering E2E tests

To trigger the E2E test for each new Release added in this PR, add a comment with the following:

`/run releases-test-suites`

If you want to trigger conformance tests, you can do so by adding a comment similar to the following:

`/run conformance-tests PROVIDER=capa RELEASE_VERSION=29.1.0`

For more details see the [README.md](/README.md#running-tests-against-prs).

PR #1421, opened 2024-09-26
Upgrade observability-bundle to 1.6.2

TheoBrigitte:
This PR upgrade the observability-bundle from 1.6.1 to 1.6.2 in CAPA v29.1.0 release.
This is to fix an issue we have with a wrong catalog name for the alloyMetrics app in the observability-bundle.
It's a very small change and would ensure we have CAPA and CAPZ v29.1.0 aligned and ready for shipping alloyMetrics.

I know this does not follow the usual release process but the change is very small and was tested.

### DIFF
--- a/capa/v29.1.0/README.md
+++ b/capa/v29.1.0/README.md
@@ -11,7 +11,7 @@
 
 - cert-exporter from v2.9.1 to v2.9.2
 - node-exporter from v1.19.0 to v1.20.0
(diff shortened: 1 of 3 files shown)


Gacko (changes requested):
We are currently working on a [CAPA v29.2.0](https://github.com/giantswarm/releases/pull/1419) which already includes this version.

Also please do not amend already released Releases. Thank you!

PR #1439, opened 2024-10-15
CAPA: Release v25.1.2.

fiunchinho:
Towards https://github.com/giantswarm/roadmap/issues/3676

### Checklist

- [ ] Roadmap issue created

### Triggering E2E tests

`/run releases-test-suites`


Gacko (changes requested):
I know, the process states using `devctl` and we didn't do this for CAPI reasons in the past (don't ask me why), but could we maybe proceed with the manual process and just copy the v25.1.1 release and change the `cluster-aws` version for now?

We can still move to `devctl` on new major releases. 🙂

Gacko:
Because it pretty much f***s up the `release.diff` and just makes it unreadable for a tiny change like this one here.

Also we should consider improving `devctl` at all, as it currently for example only lists the changelog of the app release being bumped to, but not changelogs of releases in between. So if you bump an app from v1.1.0 to v1.3.1 and there were v1.1.1, v1.2.0 and v1.3.0 in between, it just skips them leaving the user without any information about these releases.

As stated: I know, this sucks and I'd also like to have an automatic way. But currently `devctl` produces a lot less informative release and we should rather improve the tooling than missing relevant changelogs.

Gacko:
I committed an improved version, feel free to revert if you do not agree! 🙂

fiunchinho:
> I committed an improved version, feel free to revert if you do not agree! 🙂

Please, use the "add suggestion" button from the review. Otherwise it's hard to know what's going on when I try to push new changes but it fails because there are new changes that I'm not aware of.

fiunchinho:
> Because it pretty much f***s up the `release.diff` and just makes it unreadable for a tiny change like this one here.
> 
> Also we should consider improving `devctl` at all, as it currently for example only lists the changelog of the app release being bumped to, but not changelogs of releases in between. So if you bump an app from v1.1.0 to v1.3.1 and there were v1.1.1, v1.2.0 and v1.3.0 in between, it just skips them leaving the user without any information about these releases.
> 
> As stated: I know, this sucks and I'd also like to have an automatic way. But currently `devctl` produces a lot less informative release and we should rather improve the tooling than missing relevant changelogs.

Any recommended way of generating the `release.diff` file?

fiunchinho:
I tried the manual way for this one https://github.com/giantswarm/releases/pull/1442

Gacko (approved):

T-Kukawka (approved):

T-Kukawka:
@Gacko this release was needed for cluster migration and we are doing migrations to 25.1.x

Gacko:
If it's only used for migrations and not to upgrade to, fine. 🙂

PR #1460, opened 2024-10-22
CAPI: Request `cert-exporter` v2.9.2.

T-Kukawka:
Towards: https://gigantic.slack.com/archives/C6L8J93N0/p1729068032911919

### Checklist

- [ ] Roadmap issue created

### Triggering E2E tests

`/run releases-test-suites`

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,15 +1,15 @@
 releases:
+- name: "< 30.0.0"
+  requests:
+  - name: cilium
+    version: "< 0.26.0"
 - name: "> 29.1.0"
   requests:
+  - name: observability-bundle
+    version: ">= 1.6.2"
   - name: security-bundle
     version: ">= 1.8.2"
 - name: "> 29.0.0"
   requests:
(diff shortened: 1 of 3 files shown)


Gacko:
Can you also update it for CAPZ and other relevant providers? I remember this to be important because of VPA, right?

Gacko:
Also this makes CI fail because your request would expect existing releases to already have that version.

Is this really important for old releases or would it also be enough for customers to just update to a v29.x.x as soon as possible?

T-Kukawka:
@Gacko this is relevant for one of the customers as they are already facing the issue in v25 series, and they are keeping all their environments in sync, hence i would say with next release iteration it will be very beneficial to everyone to add it

T-Kukawka:
@Gacko i have added it also for vsphere, azure already has this requirement for latest and we are working with customers to update. Their workloads are not that intensive hence noone is facing this issue besides ones in CAPA v25.

Gacko (approved):

PR #1468, opened 2024-10-24
CAPI: Request `observability-bundle` >= 1.7.0.

TheoBrigitte:
Towards: https://github.com/giantswarm/roadmap/issues/3518

This PR update the releases requests to make sure we upgrade observability-bundle to v1.7.0 in newer releases

### DIFF
--- a/vsphere/requests.yaml
+++ b/vsphere/requests.yaml
@@ -3,7 +3,11 @@ releases:
   requests:
   - name: cilium
     version: "< 0.26.0"
-- name: "> 29.1.0"
+- name: ">= 29.1.0"
+  requests:
+  - name: observability-bundle
+    version: ">= 1.7.0"
+- name: "> 29.0.0"
   requests:
   - name: observability-bundle
     version: ">= 1.6.2"
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -3,6 +3,10 @@ releases:
   requests:
   - name: cilium
     version: "< 0.26.0"
+- name: ">= 29.3.0"
+  requests:
+  - name: observability-bundle
+    version: ">= 1.7.0"
 - name: "> 29.1.0"
   requests:
   - name: observability-bundle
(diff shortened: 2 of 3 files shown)


vsphere/requests.yaml
```diff
@@ -5,11 +5,13 @@ releases:
...
-  - name: observability-bundle
-    version: ">= 1.6.2"
   - name: security-bundle
     version: ">= 1.8.2"
 - name: "> 29.0.0"
   requests:
   - name: cert-exporter
     version: ">= 2.9.2"
+- name: "> 28.0.0"
+  requests:
```
QuentinBisson, line 15: I think you can't require a changé for v28

Gacko (changes requested):
We cannot introduce this in CAPV v28.0.1 nor v29.0.0. There should not be any app changes for < v29.0.0 and CAPV v29.0.0 itself should first get into alignment with what we have currently have in the latest releases of other providers.

So for CAPV, we could integrate `observability-bundle` v1.7.0 in CAPV v29.1.0 earliest.

I already talked to Nick about this and we agreed on first getting to a point where all CAPI providers use the same app versions + latest patches of Flatcar, Kubernetes and OS Tooling. This means we'll probably also forge a CAPA v29.3.1 and CAPZ v29.2.1.

Since CAPA v29.3.1 and CAPZ v29.2.1 would match the semver comparison you're introducing here and we do not ship minor app bumps in patch releases, I'd ask you to first request `observability-bundle` v1.7.0 for CAPA v29.4.0, CAPZ v29.3.0 and CAPV v29.1.0 by using `>= 29.4.0`, `>= 29.3.0` and `>= 29.1.0` as with `> 29.3.0`, `> 29.2.0` and `> 29.0.0` you would also request it for patch releases.

TheoBrigitte:
Alright I updated the requirements accordingly

vsphere/requests.yaml
TheoBrigitte: replaced this with >= 29.1.0

Gacko (approved):

PR #1478, opened 2024-10-31
Bump security-bundle version for future CAPA, CAPZ and CAPV v30 releases

fhielpos:
Bump security-bundle requests for v30 releases

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,4 +1,8 @@
 releases:
+- name: ">= 30.0.0"
+  requests:
+  - name: security-bundle
+    version: ">= 1.9.0"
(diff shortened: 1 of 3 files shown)


ssyno (approved):

stone-z (approved):

Gacko:
Does this also mean we should not or at least we do not need to bump it to v1.9.0 for a possible CAPA v29.4.0?

stone-z:
It is also fine to bump it sooner, we just aren't in a particular hurry so decided to make the request consistent for the providers.

Gacko:
Well, then you could also have requested it for the next minor releases. 🙂

PR #1480, opened 2024-10-31
CAPI: Request `security-bundle` v1.9.0.

Gacko:


Gacko:
Ehhh, I just realize there's a breaking change in v1.9.0 that probably also breaks the upgrade tests. Shouldn't it rather be a v2.0.0 then? If so, we should definitely bump this in CAPx v30.0.0 releases earliest.

/cc @giantswarm/team-shield

Gacko:
I just figured Falco is disabled by default. So all fine, go on.

fhielpos (approved):
Falco is disabled by default so we should be good 😄

PR #1488, opened 2024-11-12
CAPA: Release v29.4.0.

paurosello:
### Checklist

- [x] Roadmap issue created
- [x] Release uses latest stable Flatcar
- [x] Release uses latest Kubernetes patch version

### Triggering E2E tests

`/run releases-test-suites`

### DIFF
--- a/README.md
+++ b/README.md
@@ -12,6 +12,8 @@ to all Giant Swarm installations.
 ## AWS
 
 - v29
+  - v29.4
+    - [v29.4.0](https://github.com/giantswarm/releases/tree/master/capa/v29.4.0)
   - v29.3
     - [v29.3.1](https://github.com/giantswarm/releases/tree/master/capa/v29.3.1)
     - [v29.3.0](https://github.com/giantswarm/releases/tree/master/capa/v29.3.0)
--- a/capa/kustomization.yaml
+++ b/capa/kustomization.yaml
@@ -21,6 +21,7 @@ resources:
 - v29.2.0
 - v29.3.0
 - v29.3.1
+- v29.4.0
 
 commonAnnotations:
   giantswarm.io/docs: https://docs.giantswarm.io/use-the-api/management-api/crd/releases.release.giantswarm.io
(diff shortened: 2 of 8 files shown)


paurosello:
I don't like that we are forced to introduce a breaking change on a minor...

As phoenix we have introduced a new feature in cluster-aws which resulted in a minor release (2.4.0) and wanted to make a minor release (29.4.0) but due to the requirements from other teams we now have a breaking change.

Any thoughts? Why is there a breaking change introduced in 29.4.0 and not 30.0.0?

Gacko (changes requested):
We already have an issue open for that. Please do not just change the requests as team added their requirements there for reasons. :)

fiunchinho:
> We already have an issue open for that. Please do not just change the requests as team added their requirements there for reasons. :)

I think this was agreed on in Slack by both shield and tenet https://gigantic.slack.com/archives/C02FL4EAADD/p1731409971562999

paurosello:
> We already have an issue open for that. Please do not just change the requests as team added their requirements there for reasons. :)

I checked with both teams involved and we agreed on making this change 

https://gigantic.slack.com/archives/C02FL4EAADD/p1731409971562999

Gacko:
Can we please have the change in requests outside of this PR and also synced across all providers then? Also I'd appreciate if we could do releases with the bump in `observability-bundle` across all providers and not just CAPA. I know, this release might be important to you as you'd like to ship a feature for CAPA only, but if every team would come up with a new workload cluster release every time they release one of their apps, customers will probably never catch up with the latest releases and therefore get a consistent product experience, especially if the app versions differ across providers.

The latter probably hits us as Giant Swarm harder than the customers, as they normally only run on one provider, but this also is changing with hybrid MCs right now and so I think it's important for us who support the product to have customers use the same app versions across providers.

Apart from that we'd really appreciate to keep the content of a release in sync with previous releases. We never put a lot of details about Kubernetes or Flatcar releases into the `README.md` of CAPI releases. We usually just put links to the upstream changelog in there.

If you allow I'd like to extract the change in `requests.yaml` into a separate PR, push some suggestions to this PR and also come up with releases for CAPZ and CAPV.

paurosello:
I will put this on hold and release 29.3.1 as we need a release today to make an installation with the changes to cluster-aws only

Gacko (changes requested):
Basically LGTM, but please consider merging https://github.com/giantswarm/releases/pull/1492 first.

AndiDog (approved):

PR #1489, opened 2024-11-13
CAPA: Release v29.3.1. 

paurosello:
### Checklist

- [ ] Roadmap issue created

### Triggering E2E tests

`/run releases-test-suites`

### DIFF
--- a/README.md
+++ b/README.md
@@ -13,6 +13,7 @@ to all Giant Swarm installations.
 
 - v29
   - v29.3
+    - [v29.3.1](https://github.com/giantswarm/releases/tree/master/capa/v29.3.1)
     - [v29.3.0](https://github.com/giantswarm/releases/tree/master/capa/v29.3.0)
   - v29.2
     - [v29.2.0](https://github.com/giantswarm/releases/tree/master/capa/v29.2.0)
--- a/capa/kustomization.yaml
+++ b/capa/kustomization.yaml
@@ -20,6 +20,7 @@ resources:
 - v29.1.0
 - v29.2.0
 - v29.3.0
+- v29.3.1
 
 commonAnnotations:
   giantswarm.io/docs: https://docs.giantswarm.io/use-the-api/management-api/crd/releases.release.giantswarm.io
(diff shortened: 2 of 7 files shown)


alex-dabija (approved):

AndiDog (approved):

Gacko:
This release will definitely roll nodes and we agreed on not doing patch releases which roll nodes...

alex-dabija:
> This release will definitely roll nodes and we agreed on not doing patch releases which roll nodes...

It shouldn't. The two commits that are important between v2.3.0 and v2.4.0 of cluster-aws don't change anything for existing cluster nodes:
* [commit one](https://github.com/giantswarm/cluster-aws/commit/8285d7804c8cb01dc46b9892b9b7708596484a03#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR12) adds new tags to machine pools and the control plane if they are specified (no impact for existing clusters);
* [commit two](https://github.com/giantswarm/cluster-aws/commit/1edbb92195d852c07c9c0deb858ed45bf48fe2a4) doesn't render the subnet tags if the list is empty (it impacts only the subnets and not the nodes).
But, I might have missed something.

Could you explain why you think the release would roll the nodes? 

Yes, I agree that based on adding a new feature the release should have been a minor. We'll try do do better next time.

Gacko:
The machine template is changing. Therefore it's hash is changing. And therefore the nodes are rolling. Also you're upgrading from v2.2.0 to v2.4.0, not from v2.3.0.

Last but not least this repository requires a codeowner review, which is Tenet. So I assume you force merged this.

alex-dabija:
> The machine template is changing.

I forgot about the hash changing.

> Also you're upgrading from v2.2.0 to v2.4.0, not from v2.3.0.

Yes, correct. There are more changes that I didn't look into.

> Last but not least this repository requires a codeowner review, which is Tenet. So I assume you force merged this.

I don't know about this.

Gacko:
Also the included `cluster` chart changed and there are changes in there which might probably lead to node rolling. 🙂

PR #1508, opened 2024-11-25
CAPV: Update v29.1.0.

Gacko:


vsphere/v29.1.0/release.diff
```diff
@@ -101,4 +101,4 @@ spec:									spec:
   - name: os-tooling							  - name: os-tooling
     version: 1.20.1							    version: 1.20.1
   date: "2024-10-23T12:00:00Z"					|         date: "2024-11-13T12:00:00Z"
-  state: active								  state: active
+  state: deprecated						|         state: active
```
AverageMarcus, line 104: Curious - why wasn't the `|` there before? 🤨

AverageMarcus (approved):

vsphere/v29.1.0/release.diff
Gacko: It's a `diff --side-by-side` and the `|` indicates differences. We recently deprecated v29.0.0 but did not update the `release.diff` of v29.1.0, so from the point of view of v29.1.0 the both were still active.

Now with v29.0.0 deprecated and v29.1.0 updated, there actually is a differences as the former is deprecated and the latter still active.

PR #1543, opened 2025-01-17
delete KVM from CircleCI config

glitchcrab:
KVM is now unalived

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -24,7 +24,6 @@ workflows:
           name: filter
           mapping: |
             aws/.* push-aws true
-            kvm/.* push-kvm true
           base-revision: master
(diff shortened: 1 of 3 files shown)


njuettner (approved):
🫡

Gacko (changes requested):
Please also consider removing the `releases-kvm` in the `helm` directory.

Gacko (approved):
```sh
% grep --ignore-case kvm --recursive --exclude-dir .git --exclude-dir announcements --exclude-dir kvm --files-with-matches 
./README.md
./aws/archived/v9.0.9/README.md
./aws/archived/v9.3.6/README.md
./releases_test.go
```

I think we can keep the lines in `releases_test.go`.

PR #1544, opened 2025-01-21
CAPI: Request `security-bundle` v1.9.1.

fhielpos:
Towards: https://github.com/giantswarm/giantswarm/issues/32442

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,18 +1,12 @@
 releases:
-- name: ">= 30.0.0"
-  requests:
-  - name: security-bundle
-    version: ">= 1.9.1"
-  - name: os-tooling
-    version: ">= 1.22.1"
 - name: ">= 29.5.0"
   requests:
+  - name: cilium
+    version: ">= 0.26.0"
   - name: os-tooling
     version: ">= 1.22.1"
-- name: "< 29.5.0"
-  requests:
-  - name: cilium
-    version: "< 0.26.0"
+  - name: security-bundle
+    version: ">= 1.9.1"
 - name: ">= 29.4.0"
   requests:
   - name: coredns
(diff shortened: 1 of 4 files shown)


Gacko:
Could you please sync this across the other providers? We tried to have app versions synced as much as possible in the recent releases, so right now they mostly only differ in their provider specific apps.

Gacko (changes requested):
Could you please sync this across the other providers? We tried to have app versions synced as much as possible in the recent releases, so right now they mostly only differ in their provider specific apps.

Gacko (changes requested):
Also there is no need to add this to the `>= 30.0.0` section as `>= 29.2.0` already covers this.

And thinking about it, you should probably put this into a `> 29.2.0` as we cannot patch it in the existing v29.2.0 release.

Gacko (changes requested):
... ok, sorry for the confusion: Please request this for either the next minor or even the next major releases. IIRC v1.9.0 contains some potential breaking changes we cannot ship in patch or minor releases.

fhielpos:
Ah, true, totally forgot about that.

fhielpos:
Will check, if we need to wait for another major then there is no need in this PR.

Gacko:
If you decide to have it in a minor release, then please remove it from the `>= v30.0.0` section of all the `requests.yaml` files. There already should be a section for requesting app versions in the next minor releases. Feel free to move your request their instead.

stone-z:
The customer requesting this update is the only customer affected by the breaking change. So personally I'd be fine to even do it in a patch, but minor is fine. It doesn't need to wait for v30

Gacko:
We cannot do this in a patch, as currently and in the most recent minor we're using v1.8.x of the `security-bundle`. A minor release does not hurt, it's mostly about communication and perception. 🙂

Gacko (approved):
I updated the PR to also remove duplicates (`os-tooling` & `security-bundle` being requested for `>= 30.0.0` while already being covered by `>= 29.5.0`) and simplified the Cilium requests.

PR #1550, opened 2025-01-23
CAPA: Release v29.6.0.

AndiDog:
Towards https://github.com/giantswarm/roadmap/issues/3795, https://github.com/giantswarm/giantswarm/issues/32487, https://github.com/giantswarm/giantswarm/issues/32045

### Checklist

- [ ] Roadmap issue created


capa/v29.6.0/release.yaml
```diff
@@ -0,0 +1,136 @@
...
+    - cloud-provider-aws
+    name: aws-ebs-csi-driver
+    version: 2.30.1
+  - dependsOn:
+    - prometheus-operator-crd
+    name: aws-ebs-csi-driver-servicemonitors
+    version: 0.1.0
+  - name: aws-nth-bundle
+    version: 1.2.0-2da626ef1c0001aa4f1f38d4c9732f16bedd8a6b
+    catalog: default-test
```
AndiDog, line 20: The test catalog doesn't seem to be used. I fixed it manually in E2E clusters.

Gacko:
Is this release fixing something related to AWS Node Termination Handler? If so, please wait with releasing this. We were already about to do so or at least were talking about it. There are more things to be updated than just the AWS Node Termination Handler. Cabbage e.g. is just about to put in requests for a last Cilium patch and Shield recently requested `security-bundle` v1.9.1. So overall I'd be happy to have a talk about this and sync with releases for other providers.

Also I'd prefer using the new `devctl` approach starting with v30.

capa/v29.6.0/release.yaml
Gacko: According to the code it should be used: https://github.com/giantswarm/cluster-aws/blob/v2.5.0/helm/cluster-aws/templates/aws-nth-app.yaml#L68. This is retrieving the catalog from the release CR.

Gacko:
> Towards [...] https://github.com/giantswarm/giantswarm/issues/32045

This is not yet merged to `main` of `cluster` chart and even if it would, you could not use it as `main` of `cluster` contains breaking changes to be releases in v30 earliest.

stone-z (approved):

capa/v29.6.0/README.md
```diff
@@ -0,0 +1,88 @@
+# :zap: Giant Swarm Release v29.6.0 for CAPA :zap:
+
+<< Add description here >>
```
AndiDog, line 3: TODO

capa/v29.6.0/announcement.md
```diff
@@ -0,0 +1,3 @@
+**Workload cluster release v29.6.0 for CAPA is available**. << Add description here >>
```
AndiDog, line 1: TODO

njuettner (approved):
Structure looks good to me, can't tell about the changes in general

Gacko (changes requested):
As stated before, I'd prefer starting to use the reworked `devctl` approach in v30 earliest. It changes how the diffs look like, reorders manifests and still contains some issues around the templating in general (whitespaces, inconsistency in listing version numbers (sometimes with v, some without) and others).

Also I feel like the reduction in etcd volume size should have been shipped in a major release (and should probably also not be [back-ported to v1.3.x](https://github.com/giantswarm/cluster-aws/pull/1006) 😉).

Gacko:
Same as for v28.5.0: May I suggest merging [this PR](https://github.com/giantswarm/releases/pull/1555) on top of yours? As stated before, we'd prefer not using the new `devctl` approach for existing major releases and instead start using it in v30.0.0.

My suggestion therefore just reproduces what you already proposed by using the manual approach of copying and altering the previous minor release.

Gacko (approved):
/run releases-test-suites PREVIOUS_RELEASE=29.5.0

Gacko:
> > Also I feel like the reduction in etcd volume size should have been shipped in a major release
> 
> The storage quota (see `cluster`'s quotaBackendBytesGiB = 8) limits overall usage, so even the new default is way too high. Therefore I'd consider that okay in a minor release bump that mentions the change. We could have mentioned the rationale in the commit/changelog message, maybe.

You're absolutely right on that. From a Giant Swarm and technical point of view, the reduction probably doesn't hurt. But I don't know the customer / user point of view on this. So let's just hope nobody's complaining about it in the end. 😅

> so even the new default is way too high

One note on that: Customers can configure the quota to 16 GB. Worst case and depending on how etcd is defragmenting / rewriting its database files and from my experience with databases, I'd expect this to consume twice the amount of the quota. So 32 GB. We can still further reduce it, but I'd not save in the wrong place.

AndiDog (commented):
@giantswarm/team-cabbage could you also take a look? Particularly if the Cilium bump is fine?

ubergesundheit (approved):
cilium-wise LGTM

T-Kukawka:
added release notes, please review

PR #1551, opened 2025-01-23
CAPI: Request Cilium v0.25.2.

ubergesundheit:
This PR requests  cilium-app 0.25.2 with v1.15.13 on azure, capa, cloud-director and vsphere


Gacko (changes requested):
Can you please elaborate on your thoughts and constraints so we can maybe better match the requests with our plans for upcoming releases?

v25 to v28 releases are meant to be all the same without any changes in app versions, if possible. So the earliest release we can introduce such changes on is v29.

Inside of v29 it unlikely we will have patch releases on older minors as we normally ask customers to just go to the latest minor.

Additionally Kubernetes v1.29.x is already no longer maintained or at least about to be discontinued soon.

So after all I'd ask you to just request this Cilium app version for any release after the most recent v29 minor release of a provider to not make the `requests.yaml` unnecessarily complex.

And IIRC we wanted to go for Cilium v0.26.0+ on v30.

Gacko:
Also please avoid using constraints like `< 0.25.99` and rather use `< 0.26.0` instead.

Gacko:
Updated my comment.

Gacko:
> can you explain this one?

A `0.25.100` would not match the expression while still being valid, right?

azure/requests.yaml
```diff
@@ -27,3 +27,7 @@ releases:
   requests:
   - name: cert-exporter
     version: ">= 2.9.2"
+- name: "> 25.0.1 < 26.0.0 || > 26.0.1 < 27.0.0 || > 27.0.1 < 28.0.0 || > 28.0.1 < 29.0.0"
```
Gacko: This would first affect v25.0.2, v26.0.2 and so on. Also, what about v29 releases? Don't they need Cilium app v0.25.2?

ubergesundheit:
> Can you please elaborate on your thoughts and constraints so we can maybe better match the requests with our plans for upcoming releases?

Include cilium-app v0.25.2 everywhere where v0.25.1 is used atm except for release 30.0.0 because we want to make the transition to cilium v1.16 (cilium-app >= v0.26.0) as smoth as possible. v0.25.2 contains the latest patch release of cilium

capa/requests.yaml
```diff
@@ -31,3 +31,7 @@ releases:
   requests:
   - name: cert-exporter
     version: ">= 2.9.2"
+- name: "> 25.1.4 < 25.2.0 || > 25.2.2 < 25.3.0 || > 25.3.1 < 25.4.0 || > 25.4.1 < 26.0.0 || > 26.1.2 < 26.2.0 || > 26.2.1 < 26.3.0 || > 26.3.1 < 27.0.0 || > 27.1.2 < 27.2.0 || > 27.2.1 < 27.3.0 || > 28.1.3 < 28.2.0 || > 28.2.1 < 28.3.0 || > 28.3.1 < 28.4.0 || > 28.4.1 < 29.0.0 || > 29.1.1 < 29.2.0 || > 29.2.1 < 29.3.0 || > 29.3.1 < 29.4.0 || > 29.4.1 < 29.5.0 || > 29.5.1 < 29.6.0"
```
Gacko: Same here: This would first affect v25.1.5, v25.2.3 and so on.

As already noted before, I'd prefer not requesting a Cilium update for these versions and just do it for v29.

Gacko (commented):
From re-reading your proposal, I think what you want to achieve is having one last patch of Cilium before going to Cilium v0.26.0 and/or any v30 release of our platform, right?

Gacko:
Ok, got it. I'd suggest requesting v0.25.2 for any new v29 release then and at the same time make sure v0.26.0 is used starting from v30. If you agree, I can propose a change to your PR.

ubergesundheit:
Yes please

Gacko:
Done!

ubergesundheit:
thanks :bow:

Gacko (approved):

PR #1591, opened 2025-02-19
CAPI: Fix Cilium requests.

Gacko:
v30.0.0 matches `>= 29.5.0`, but requests Cilium `>= 0.26.0`. This clashes and as we already fulfill the old request, I'm going to remove the upper bound.

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -6,7 +6,7 @@ releases:
 - name: ">= 29.5.0"
   requests:
   - name: cilium
-    version: ">= 0.25.2 < 0.26.0"
+    version: ">= 0.25.2"
   - name: os-tooling
     version: ">= 1.22.1"
   - name: security-bundle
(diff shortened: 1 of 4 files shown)


azure/requests.yaml
```diff
@@ -6,7 +6,7 @@ releases:
 - name: ">= 29.5.0"
   requests:
   - name: cilium
-    version: ">= 0.25.2 < 0.26.0"
```
mcharriere, line 9: this was meant to guarantee that no v29.x would ship 0.26.

Gacko: I am aware of this. We already have a v29.x.x out with Cilium v0.25.2 and there won't be any more v29.x.x releases as we are currently working on v30.0.0.

So technically this is not perfect and you are right, but fixing it properly would make the requests more complex to read.

Gacko: I heavily doubt anyone will bump Cilium to anything behind v0.25.x in a new v29.x.x release, if we'd still make one.

mcharriere: > there won't be any more v29.x.x releases

I don't buy that :)

Gacko: Me neither, but let's hope. 😬

PR #1605, opened 2025-03-10
Request nvidia_runtime be included from release v30.1 onwards

AverageMarcus:
Requesting that the os-tooling be updated to at least v1.24.0 starting from release v30.1 onwards which will include the nvidia_runtime binaries and config on the node image.

Closes: https://github.com/giantswarm/roadmap/issues/3886

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,4 +1,8 @@
 releases:
+- name: ">= 30.1.0"
+  requests:
+  - name: os-tooling
+    version: ">= 1.24.0"
 - name: ">= 30.0.0"
   requests:
   - name: cilium
--- a/capa/requests.yaml
+++ b/capa/requests.yaml
@@ -1,4 +1,8 @@
 releases:
+- name: ">= 30.1.0"
+  requests:
+  - name: os-tooling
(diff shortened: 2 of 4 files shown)


Gacko:
As noted on Slack, this should not be required, as we usually pull in the latest Kubernetes and Flatcar versions anyway when doing a new minor. As these are always built using the latest CAPI Image Builder version, they already include the NVIDIA runtime.

AverageMarcus:
> we usually pull in the latest Kubernetes and Flatcar versions anyway 

But that's not enforced, right? What's to stop a new minor that just includes a new (for example) `cluster-aws` version because it causes node rolling?

> As these are always built using the latest CAPI Image Builder version

The latest Kubernetes / Flatcar versions also exist with the previous version of the os-tooling though, so if someone forgets to bump this value too then we wont get the nvidia_runtime.

I guess I don't really understand the point of the requests then. We don't have any automation yet for creating a new release so this is the only safeguard we have to make sure required things are included, right? 

(I created this PR because that's what I was informed to do last time we needed a new tooling release - https://github.com/giantswarm/releases/pull/1538)

Gacko (approved):

PR #1606, opened 2025-03-11
Update cilium requests

mcharriere:
- **Update requested version for cilium-app**
- **Add requested version for future releases**

### Checklist

- [ ] Roadmap issue created

### Triggering E2E tests

`/run releases-test-suites`

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,4 +1,8 @@
 releases:
+- name: "31.x.x"
+  requests:
+  - name: cilium
+    version: ">= 0.32.0"
 - name: ">= 30.1.0"
   requests:
   - name: os-tooling
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -3,10 +3,10 @@ releases:
   requests:
   - name: os-tooling
     version: ">= 1.24.0"
-- name: ">= 30.0.0"
+- name: "30.x.x"
   requests:
   - name: cilium
-    version: ">= 0.26.0"
+    version: "0.31.x"
 - name: ">= 29.5.0"
   requests:
   - name: cilium
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -7,7 +7,7 @@ releases:
   requests:
   - name: os-tooling
     version: ">= 1.24.0"
-- name: ">= 30.0.0"
+- name: "30.x.x"
   requests:
   - name: cilium
     version: "0.31.x"
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,5 +1,5 @@
 releases:
-- name: "31.x.x"
+- name: ">= 31.0.0"
   requests:
   - name: cilium
     version: ">= 0.32.0"
@@ -7,7 +7,7 @@ releases:
   requests:
   - name: os-tooling
     version: ">= 1.24.0"
-- name: "30.x.x"
+- name: ">= 30.0.0"
   requests:
   - name: cilium
     version: "0.31.x"
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -7,7 +7,7 @@ releases:
   requests:
   - name: os-tooling
     version: ">= 1.24.0"
-- name: "30.x.x"
+- name: ">= 30.0.0 < 31.0.0"
   requests:
   - name: cilium
     version: "0.31.x"
--- a/capa/requests.yaml
+++ b/capa/requests.yaml
@@ -1,4 +1,8 @@
 releases:
+- name: "31.x.x"
+  requests:
+  - name: cilium
+    version: ">= 0.32.0"
 - name: ">= 30.1.0"
   requests:
   - name: os-tooling
--- a/capa/requests.yaml
+++ b/capa/requests.yaml
@@ -7,12 +7,14 @@ releases:
   requests:
   - name: os-tooling
     version: ">= 1.24.0"
+- name: "30.x.x"
+  requests:
+  - name: cilium
+    version: "0.31.x"
 - name: ">= 30.0.0"
   requests:
   - name: aws-ebs-csi-driver
     version: ">= 3.0.0"
-  - name: cilium
-    version: "0.31.x"
 - name: ">= 29.6.0"
   requests:
   - name: aws-pod-identity-webhook
--- a/capa/requests.yaml
+++ b/capa/requests.yaml
@@ -3,12 +3,14 @@ releases:
   requests:
   - name: os-tooling
     version: ">= 1.24.0"
+- name: "30.x.x"
+  requests:
+  - name: cilium
+    version: "0.31.x"
 - name: ">= 30.0.0"
   requests:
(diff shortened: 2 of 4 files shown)


Gacko (changes requested):
Can we please stick to semver comparisons? Because defining major releases without operators will fail the moment we do a v32.0.0.

Doing a v31.0.0 with Cilium v0.32.0 matches both requirements.

mcharriere:
> Can we please stick to semver comparisons? Because defining major releases without operators will fail the moment we do a v32.0.0.

I'm not sure what you mean here. Comparing semver with `x` is valid, easier to understand and more compact. 

> Doing a v31.0.0 with Cilium v0.32.0 matches both requirements.

Which requirements?

Gacko:
> I'm not sure what you mean here. Comparing semver with `x` is valid, easier to understand and more compact.

Sure, it is. But it pins requests to single releases. Using greater than or equal keeps these requirements compatible with later releases and does not limit them to only v30.x.x or v31.x.x.

Other teams might want their requests to match more than one workload cluster release and would therefore either need to put in a semver comparison like yours for each possible workload cluster release or use `>=` alongside with your `30.x.x` comparison. Either way it's duplicate code.

Using `>=` improves compatibility, makes the file less complex to understand and does not require requesting specific versions for each possible workload cluster release.

mcharriere:
> Sure, it is. But it pins requests to single releases. Using greater than or equal keeps these requirements compatible with later releases and does not limit them to only v30.x.x or v31.x.x.

Doing it with `>=`  does not actually represent what we need. For that we should set a range, making it less clearer and prone to errors. 

> Other teams might want their requests to match more than one workload cluster release and would therefore either need to put in a semver comparison like yours for each possible workload cluster release or use >= alongside with your 30.x.x comparison. Either way it's duplicate code.

This is not code, this is a system of **declarative** constrains. Trying to reduce the verbosity of constrains only complicates things. 

> Using >= improves compatibility, makes the file less complex to understand and does not require requesting specific versions for each possible workload cluster release.

using `>=` makes things more complicated when trying to express what we want to express here. 
Take a look at the PR now, it might look simpler, but it doesn't actually work. The check is evaluating rule by rule, being the older constrains evaluated later. This forces us to do things like using v19.99.99 as we've seen in the past.

> You're requesting Cilium to be >= v0.31.0 for Releases >= v30.0.0 and Cilium to be >= v0.32.0 for Releases >= v31.0.0.

No, this is not what we want to express. we need cilium v0.31.x in v30 and cilium >= 0.32.0 in v31. 

> Both of these requirements are met if we use Cilium v0.32.0 in Release v31.0.0.

Then we are not establishing the requirement.

> Overall I do not see a benefit of pinning requests to major releases by using this syntax.

The benefit is that it's the easiest way to express what we need to express without resorting on ranges. We could use `^` or `~`; but IMO those are less understood than the `x` as wildcard.

Gacko:
Hey!

Thank you very much for your detailed feedback!

First of all the current state of the change you're requesting looks good to me. I think from a product point of view it's a solid solution to have the platform releases use ranges (`name: ">= 31.0.0"`) while being free on what to use in the app releases part `version: "0.31.x"`.

This leaves you as a provider of an app the freedom to precisely decide which version to use while, on the other hand, the people defining platform releases do not need to take care of requests not matching any longer when doing a new, let's say, major release, because the app team only requested `name: "30.x.x"` instead of `name: ">= 30.0.0"`.

I heavily doubt this to become an issue, but with using constraints like `name: "30.x.x"` you might - for whatever reason - potentially not bump the app when doing a platform release v31.0.0, as long as there is no explicit `name: "31.x.x"` in the requests. So using `name: ">= 30.0.0"` is just a safer way of expressing to use a specific version IMHO, because it includes v30.0.0 and **above instead** of only v30.0.0.

Nothing prevents you from still putting in a request for Cilium v0.32.0 in platform releases v31.0.0, though. So in the end both of our requirements are met: Yours to have at least Cilium v0.31.0 in platform release v30.0.0 and v0.32.0 in v31.0.0 and ours to not miss any requirements just because they were defined using a `name: "30.x.x"` pattern.

Gacko:
So from going through [the code](https://github.com/giantswarm/releases/blob/master/releases_test.go#L334-L368), we're doing the following:

1. Gather all the requests matching a release. So for a v31.0.0 there would be a request for Cilium to be v0.31.x and a request for Cilium to be >= v0.32.0.
2. Check each request against the apps listed in the release by verifying that at least one app matches the request.
3. Fail is none of the apps satisfies a given request.

With the current state of the code, we would make sure any v30.x.x uses exactly Cilium v0.31.x, because the requests for v31.0.0 are not being considered for a v30.x.x release.

For a v31.x.x release, both requests would be considered, but using Cilium v0.31.x would not match the request for Cilium >= v0.32.0. Therefore at least on request is not satisfied and the whole test fails.

I can put in draft PRs to verify this if you want.

vsphere/requests.yaml
```diff
@@ -1,12 +1,16 @@
...
+- name: ">= 31.0.0"
+  requests:
+  - name: cilium
+    version: ">= 0.32.0"
 - name: ">= 30.1.0"
   requests:
   - name: os-tooling
     version: ">= 1.24.0"
-- name: ">= 30.0.0"
+- name: "30.x.x"
```
Gacko, line 10:
```suggestion
- name: ">= 30.0.0 < 31.0.0"
```

cloud-director/requests.yaml
```diff
@@ -1,12 +1,16 @@
...
+- name: ">= 31.0.0"
+  requests:
+  - name: cilium
+    version: ">= 0.32.0"
 - name: ">= 30.1.0"
   requests:
   - name: os-tooling
     version: ">= 1.24.0"
-- name: ">= 30.0.0"
+- name: "30.x.x"
```
Gacko, line 10:
```suggestion
- name: ">= 30.0.0 < 31.0.0"
```

capa/requests.yaml
```diff
@@ -1,14 +1,20 @@
 releases:
+- name: ">= 31.0.0"
+  requests:
+  - name: cilium
+    version: ">= 0.32.0"
 - name: ">= 30.1.0"
   requests:
   - name: os-tooling
     version: ">= 1.24.0"
+- name: "30.x.x"
```
Gacko, line 10:
```suggestion
- name: ">= 30.0.0 < 31.0.0"
```

azure/requests.yaml
```diff
@@ -1,12 +1,16 @@
...
+- name: ">= 31.0.0"
+  requests:
+  - name: cilium
+    version: ">= 0.32.0"
 - name: ">= 30.1.0"
   requests:
   - name: os-tooling
     version: ">= 1.24.0"
-- name: ">= 30.0.0"
+- name: "30.x.x"
```
Gacko, line 10:
```suggestion
- name: ">= 30.0.0 < 31.0.0"
```

Gacko (commented):
As discussed on Slack, some suggestions for readability.

PR #1655, opened 2025-04-24
CAPA: Request `cert-manager-crossplane-resources` v0.1.0.

ssyno:
### Checklist

- [ ] Roadmap issue created

### Triggering E2E tests

`/run releases-test-suites`

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -3,8 +3,6 @@ releases:
   requests:
   - name: cilium
     version: ">= 0.32.0"
-  - name: cert-manager-crossplane-resources
-    version: ">= 0.1.0"
 - name: ">= 30.1.0"
   requests:
   - name: os-tooling
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -3,6 +3,8 @@ releases:
   requests:
   - name: cilium
     version: ">= 0.32.0"
+  - name: cert-manager-crossplane-resources
+    version: ">= 0.1.0"
 - name: ">= 30.1.0"
   requests:
   - name: os-tooling
--- a/capa/requests.yaml
+++ b/capa/requests.yaml
@@ -2,7 +2,7 @@ releases:
 - name: ">= 31.0.0"
   requests:
   - name: cert-manager-crossplane-resources
-    version: ">= 0.0.7"
+    version: ">= 0.1.0"
   - name: cilium
     version: ">= 0.32.0"
 - name: ">= 30.1.0"


azure/requests.yaml
```diff
@@ -3,6 +3,8 @@ releases:
   requests:
   - name: cilium
     version: ">= 0.32.0"
+  - name: cert-manager-crossplane-resources
```
Gacko, line 6: This app is not a part of `cluster-azure`.

Gacko (changes requested):

azure/requests.yaml
Gacko: Then please request it as soon as it is. Otherwise this blocks release creation.

Also then the app should have been integrated in the `cluster` chart instead of just `cluster-aws`, so the way it's getting deployed is the same across providers. Config can still differ via using Helm templates.

Gacko (changes requested):

azure/requests.yaml
ssyno: Sure, I converted this into a draft and I will make it ready for review once its ready.
This app can't be integrated into the `cluster` chart for now as the way its getting handled in each provider is different

Gacko: I assume you're talking about the way it's being configured, right? Not the way the app CR looks like? Because we considered this case in `cluster` chart. You can have an app in the `cluster` chart and still decide if it's being installed at all or how it's being configured in the provider specific chart. This is what we have the `providerIntegration` values for.

Check how `coredns` is being implemented in the `cluster` chart and how it is being integrated in `cluster-aws` for example:

https://github.com/giantswarm/cluster/blob/main/helm/cluster/files/helmreleases/coredns.yaml
https://github.com/giantswarm/cluster-aws/blob/main/helm/cluster-aws/templates/_coredns_helmrelease_config.yaml

Gacko: Same goes for Cilium: It's an app in `cluster` chart, but config comes from provider specific cluster chart.

Gacko: So as long as their is nothing provider specific in [this part](https://github.com/giantswarm/cluster-aws/blob/main/helm/cluster-aws/templates/cert-manager-crossplane-resources-app.yaml#L36-L65) of the app, there's no reason not to have it in the `cluster` chart.

Gacko (approved):

azure/requests.yaml
ssyno: I removed it from `Azure` releases, as `DNS01` `clusterIssuer` is already supported in `CAPZ` private clusters with a different approach. Lets go for the minor release for CAPA as there are breaking changes in it.

PR #1657, opened 2025-04-28
Archive unused releases

yulianedyalkova:
Archive most unused releases for CAPZ, CAPV and CAPVCD based on [this](https://giantswarm.grafana.net/d/be9a0bh8mbwn4e/capi-releases?orgId=1&from=now-6h&to=now&timezone=browser&var-installation=$__all&var-provider=capa&var-repo=capa&var-release_version=$__all&var-customer=$__all) dashboard.

### Checklist

- [ ] Roadmap issue created

### Triggering E2E tests

`/run releases-test-suites`


Gacko (approved):

Gacko:
CI is failing due to some Python incompatibility in the `circleci/path-filtering` orb. We didn't change this in ages, it just broke out of nowhere. Recent commits were all good. I think CircleCI changed something under the hood.

Anyway. We only need this action for checking if there have been changes in the Vintage releases directories. As of now this is only `aws/` left. As long as we are no longer releasing Vintage AWS, we can also just remove the whole action and everything related.

WDYT?

yulianedyalkova:
If we don't fix it, the next person creating a PR would hit it so would rather fix now :) And it seems like we've been skipping CI on all PRs for the past 2 weeks!?

Gacko:
No, this part of CI isn't skipped by the label.

Also what I suggest isn't fixing the check but removing its necessity.

yulianedyalkova:
The issue is upstream and they're working on fixing it: https://support.circleci.com/hc/en-us/articles/36392688494747-Python-3-8-is-not-supported-on-the-path-filtering-orb so can actually ignore in this case.

Gacko (approved):

PR #1668, opened 2025-05-08
Add release deprecation automation

njuettner:
Towards: https://github.com/giantswarm/roadmap/issues/3988

This PR adds an automated GitHub workflow that deprecates outdated releases.

## How It Works

The workflow applies these deprecation rules:

- When a new minor/patch release is published, predecessor versions are marked as deprecated
- Releases actively used by customer clusters remain active
- After a new major version, releases from two major versions back are deprecated (if not needed as upgrade paths)

The workflow runs **weekly** and can also be **triggered manually**.

## Example PR Output

image

## Implementation PRs

- #1679
- #1680 (bigger change 😬)
- #1681
- #1682

## Notes

- Each provider gets its own PR to keep it clean
- Changes currently only `release.yaml` files and provider-level `releases.json`, I'd like to avoid the `diff` files 😬 that would be painful.


.github/workflows/deprecate-releases.yaml
```diff
@@ -0,0 +1,155 @@
...
+  # Run weekly on Monday morning
+  schedule:
+    - cron: "0 6 * * 1" # Every Monday at 6:00 AM UTC
+
+  # Allow manual triggering
+  workflow_dispatch:
+  # For testing only
+  push:
+    paths:
+      - ".github/workflows/deprecate-releases.yaml"
```
njuettner, line 13: This is only for testing, once approved I'll remove that.

tools/deprecate-release.go
```diff
@@ -0,0 +1,429 @@
...
+		}
+	}
+}
+
+// checkReleasesInUse queries Grafana for all active releases in a single API call
+func checkReleasesInUse(releases []*ReleaseInfo, apiKey string, provider string, verbose bool) {
+	// Convert provider name for API if needed
+	apiProvider := provider
+	if provider == "azure" {
+		apiProvider = "capz"
```
njuettner, line 155: Grafana provider name needs to rewired for Azure

.github/workflows/deprecate-releases.yaml
```diff
@@ -0,0 +1,224 @@
...
+          fi
+
+      - name: Create Pull Request
+        if: ${{ steps.changed-files.outputs.has_changes == 'true' }}
+        uses: peter-evans/create-pull-request@v5
+        with:
+          token: ${{ secrets.GITHUB_TOKEN }}
+          commit-message: "${{ steps.provider-name.outputs.display_name }}: Deprecate outdated releases"
+          title: "${{ steps.provider-name.outputs.display_name }}: Deprecate outdated releases"
+          draft: true
```
njuettner, line 204: Once approved I will change it

AverageMarcus (changes requested):
Looks really good to me, just some small comments. :) 

One thing I couldn't see but might have missed - how do we handle the "keep old versions to allow upgrade path for v25" ?

.github/workflows/deprecate-releases.yaml
```diff
@@ -0,0 +1,221 @@
...
+            echo "EOF" >> $GITHUB_OUTPUT
+            echo "has_changes=true" >> $GITHUB_OUTPUT
+          else
+            echo "No release files were changed"
+            echo "has_changes=false" >> $GITHUB_OUTPUT
+          fi
+
+      - name: Create Pull Request
+        if: ${{ steps.changed-files.outputs.has_changes == 'true' }}
+        uses: peter-evans/create-pull-request@v5
```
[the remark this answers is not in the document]
njuettner: There's no official one. I pinned it.

njuettner:
@AverageMarcus that is covered in the `deprecate-releases.go`file, we keep the upgrade path as long as we have certain older releases in `use`. We'll keep the latest release of each major until the customer has upgraded.

.github/workflows/deprecate-releases.yaml
```diff
@@ -0,0 +1,216 @@
...
+            *)
+              echo "display_name=${PROVIDER^^}" >> $GITHUB_OUTPUT
+              ;;
+          esac
+
+      - name: Build deprecation tool
+        run: |
+          cd tools
+          go mod tidy
+          go build -o deprecate-releases tools/deprecate-release.go
```
AverageMarcus: Not sure `tools/` is needed here is it? We're already in the tools dir

njuettner: Yeah I noticed it :D I removed it

AverageMarcus (approved):

.github/workflows/deprecate-releases.yaml
AverageMarcus: You're too quick for me! 😆

PR #1701, opened 2025-05-27
Adding functionality to set `deprecated` status in  `*.diff` files

njuettner:
Towards: [#33509](https://github.com/giantswarm/giantswarm/issues/33509)

Added functionality to handle the `release.diff` files when a release is deprecated. 

1.  **Correct Status Display:**
    *   The script now makes sure the `state:` (like `active` or `deprecated`) shown in the `release.diff` file is correct for *both* the release we're marking as old and the one just before it.
    *   If the previous release is still `active`, the diff will say `state: active` on the left but if it's already `deprecated` it will change it.
    *   The release we're currently marking as old will always say `state: deprecated` on the right.

2.  **Fixes Lines Missing the `|` Separator:**
    *   Sometimes, the `state:` line in the diff looked like `state: active     state: active` (no `|` in the middle).
    *   If the statuses need to be updated (e.g., the right side needs to become `deprecated`), the script will now fix this line to look like `state: active | state: deprecated`.
    *   If the statuses on such a line are already correct and don't need changing, the script will leave it as is.

Example:

CAPA `v29.6.1` is deprecated, if `v29.6.2` would become deprecated. The `release.diff` might look like:

`state: active                                 state: active`

it would be changed to:

`  state: deprecated                     |         state: deprecated`


Gacko:
I might be a little late, but here's what I'm currently using to just update the diffs of all releases:

```sh
for release in {capa,azure,vsphere,cloud-director}/v*.*.*
do                
  previous="${release:h}/v$(grep "compared to" "${release}/README.md" | head -n 1 | grep --extended-regexp "\d+\.\d+\.\d+" --only-matching)"
  [ -f "${previous}/release.yaml" ] || previous="${previous:h}/archived/${previous:t}"
  echo "# ${previous} <> ${release}"
  diff --side-by-side "${previous}/release.yaml" "${release}/release.yaml" > "${release}/release.diff"
done
```

Gacko:
Also I saw we are currently updating the `releases.json` by manipulating the JSON file. Couldn't we just run what's in the `renovate.go`?

Gacko:
I thought the field is already getting changed in the current automation? Like, changing the `release.yaml` as it is right now is totally fine, as YAML is structured. But the `release.diff` is mostly human-readable and meant for them. Normally, when I'm doing releases, I create them by running...

```sh
diff --side-by-side "${previous}/release.yaml" "${release}/release.yaml" > "${release}/release.diff"
```

... where `previous` is the release the new release is based on. This is then also being noted in the `Changes compared to` part. And this is what my script is then relying on. So no `Changed compared to`, no script. 😬

PR #1741, opened 2025-06-19
Fix catalog

fiunchinho:
This was over looked during the review.


Gacko:
You cannot simply amend already released Releases, because we also create GitHub Releases from them and use the uploaded resources in MC Bootstrap.

PR #1753, opened 2025-06-26
CAPI: Request Flatcar < v4230.2.0.

Gacko:
We'd like to prevent Flatcar from being bumped to v4230.2.0 and beyond in any release before v32.0.0 because it deprecates cgroups v1 and one of our customers is still using them. This change in requests makes sure they can still use v31.0.0 and have enough time to migrate.

We can still adjust these requests and the used Flatcar version in case there are some security issues only being fixed in v4230.2.0 or later, but let's start from here so nobody accidentally puts in an "incompatible" version. 🙂

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -1,4 +1,8 @@
 releases:
+- name: "< 32.0.0"
+  requests:
+  - name: flatcar
+    version: "< 4230.2.0"
 - name: ">= 31.0.0"
   requests:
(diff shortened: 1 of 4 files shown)


Gacko:
This has been tested in this draft PR: https://github.com/giantswarm/releases/pull/1751.

yulianedyalkova (approved):

Gacko:

PR #2100, opened 2025-12-17
Fix release deprecation workflow to update and keep diff files

njuettner:
This PR fixes two issues preventing `release.diff` files from being updated correctly during automated release deprecation:

1.  **Workflow Reversion:** The workflow was previously configured to discard changes to `release.diff` files. This has been updated to include `release.diff` in the list of kept files.
2.  **Tool Logic:** The `deprecate-release` tool now iterates over active releases to update their diff files when a previous release is deprecated (handling the "left side" state change in the diff). It also correctly updates the in-memory state of the deprecated release so subsequent checks are accurate.

Fixes #2098 (related context).

### DIFF
--- a/.github/workflows/archive-releases.yaml
+++ b/.github/workflows/archive-releases.yaml
@@ -48,6 +48,10 @@ jobs:
         run: |
           set -eo pipefail
 
+          # Create a temporary directory for all temp files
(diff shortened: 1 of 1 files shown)


Gacko:
Thank you! 🙏 Yet another thing I always just did manually instead of fixing the root cause. 🥲

PR #2104, opened 2025-12-17
Fix GH action for announcing releases

njuettner:


.github/workflows/announce-release.yaml
```diff
@@ -64,25 +65,31 @@ jobs:
...
     outputs:
-      announcement: ${{ steps.get_announcement.outputs.announcement }}
+      announcement: ${{ steps.get_announcement_md.outputs.announcement }}
     steps:
       - name: Get announcement.md
         id: get_announcement_md
         run: |
-          export LINK="https://raw.githubusercontent.com/giantswarm/releases/refs/heads/master/${{ inputs.provider }}/${{ inputs.version }}/announcement.md"
+          PROVIDER=${{ inputs.provider }}
+          if [[ "$PROVIDER" == "capz" ]]; then
```
njuettner: if selecting capz as the provider we must convert it to azure to find the right directory path

salaxander (approved):

PR #2117, opened 2026-01-13
Request observability-bundle 2.5.0 on next releases

TheoBrigitte:
This PR requests the new observability-bundle `v2.5.0` to be part of the upcoming releases on all providers.

### DIFF
--- a/azure/requests.yaml
+++ b/azure/requests.yaml
@@ -7,6 +7,8 @@ releases:
   requests:
   - name: cilium
     version: ">= 1.3.0"
+  - name: observability-bundle
+    version: ">= 2.5.0"
 - name: "< 32.0.0"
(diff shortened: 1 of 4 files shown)


Gacko:
Also your change would break existing releases.

Gacko:
Sorry, I think my message was a little rude. Just wanted to tell there's no real need for such PRs and we would request your review for new releases anyway. So in case we forget about a version, you are still being notified and get the chance to check.

Regarding the broken releases: This does not affect the current v34.0.0 PR, but since you requested it for `>= 32.0.0`, it would also get requested for v33.x.x. But since these releases already exist and cannot get amended, the CI fails since your requested cannot be fulfilled. Does that make sense? 😅

PR #2253, opened 2026-04-14
Include Release CR in preview charts with SHA-suffixed name

njuettner:
PR preview charts are published with SHA-suffixed versions (e.g. `35.0.0-<sha>`),
but the Release CR on the cluster keeps its clean name (`aws-35.0.0`). The cluster
chart's lookup function (`cluster.internal.get-release-resource`) constructs the
Release CR name from `global.release.version`, so it tries to find `aws-35.0.0-<sha>`
and fails.

Two changes:
- Set `global.release.version` to the SHA-suffixed version in preview builds
- Copy the Release CR into the chart's templates with the patched name, so
  the lookup finds it when the chart is installed


njuettner:
Ignore that :D

PR #2271, opened 2026-04-22
clean old requests, request security-bundle bump

stone-z:
Cleans up some old requests and adds requests for security-bundle bumps in potential v33 and v34 patches.

### DIFF
--- a/capa/requests.yaml
+++ b/capa/requests.yaml
@@ -23,4 +23,63 @@ releases:
   requests:
   - name: cilium
     version: ">= 1.2.0 < 1.3.0"
-
+- name: ">= 31.1.0"
+  requests:
+  - name: aws-nth-bundle
+    version: ">= 1.2.2"
+  - name: coredns
+    version: ">= 1.26.0"
+  - name: k8s-dns-node-cache
+    version: ">= 2.9.0"
+- name: ">= 31.0.0"
+  requests:
+  - name: cert-manager-crossplane-resources
+    version: ">= 0.1.0"
+- name: ">= 30.1.2 < 31.0.0"
+  requests:
+  - name: cilium
+    version: ">= 0.31.4"
+- name: ">= 30.1.0"
+  requests:
+  - name: os-tooling
+    version: ">= 1.24.0"
+  - name: security-bundle
+    version: ">= 1.10.0"
+- name: ">= 30.0.0 < 30.1.2"
+  requests:
+  - name: cilium
+    version: "0.31.x"
+- name: ">= 30.0.0"
+  requests:
+  - name: aws-ebs-csi-driver
+    version: ">= 3.0.0"
+- name: ">= 29.6.0"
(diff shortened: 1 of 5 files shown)


njuettner (approved):

Gacko (changes requested):
Please do not remove old requests. I know, it's not likely we gonna need them again, but as long as they don't hurt, I'd rather keep them since they are kind of tests for our Releases.

Also please request the bump across all providers. Thank you! 🙂

stone-z:
.... so never clean them up?

capa/requests.yaml
```diff
@@ -3,6 +3,14 @@ releases:
   requests:
   - name: cluster-aws
     version: "< 7.0.0"
+- name: ">= 34.2.0 < 35.0.0"
+  requests:
+  - name: security-bundle
+    version: ">= 1.17.1 < 1.18.0"
```
Gacko, line 5: Can you maybe move this up a bit? Above the `< 34.0.0` constraint, please. 🙂

capa/requests.yaml
```diff
@@ -3,6 +3,14 @@ releases:
...
+- name: ">= 33.3.0 < 34.0.0"
```
Gacko, line 10: What about patches to v33.2.0? 😅 I guess `> 33.2.0 < 34.0.0` would be better here, right?

Gacko (changes requested):
Well, basically you're right, we could clean them up at some point. But it does not hurt and I don't know about the future, so I'd rather keep them as they define which versions to use in which release and if someone for some weird reason wants to create a release for those old versions...

capa/requests.yaml
stone-z: Semantically, it updates several components which have new features for customers in a backwards-compatible way. We've been asked to do that in minors, but I can do it in a patch if you prefer

stone-z (commented):
When I [designed this feature](https://github.com/giantswarm/releases/pull/373) ( :wink: ) the understanding was that a future release would "never" downgrade the version of a given component, except for special intentional cases. So if a release has already been cut in each past minor matched by a request in a given provider, it was safe to remove the request. In fact we planned to drop them automatically, but never got around to it.

I've expanded the requests to the other providers

stone-z:
I can also bring back the old requests if you'd like

capa/requests.yaml
Gacko: No, all good. Was just wondering since you're requesting a specific patch level and so I'd assume this bump to be safe for patch releases.

Gacko (commented):
I'd say keep them (=> do not remove them) for now and leave an issue for Tenet, so we can think about how to proceed here, because ideally we do not clean them up manually but in conjunction with the deprecation & archival workflow we already set up.

Gacko (approved):

Gacko:

PR #2267, opened 2026-04-21
CAPVCD: Release v33.3.0.

taylorbot:

PR #2274, opened 2026-04-23
Fix duplicate release announcements in Slack

njuettner:
Fixes duplicate Slack announcements when a deprecate/archive PR merges shortly after a release PR.

**Cause**: `notify-news-product` announced any GitHub release whose `createdAt` was within a 30-minute wall-clock window. Every push to master that touched a `release.yaml` re-ran the job, and deprecate/archive PRs (which modify existing `release.yaml` files) re-announced the still-fresh release. This produced the 3x announcement seen for VMware Cloud Director v33.3.0 (PR #2267 → #2268 → #2269, all within 16 minutes).

**Fix**: Announce only releases whose `release.yaml` was *added* in the commits of the triggering push (`git diff --diff-filter=A "${before}..${sha}"`, with a `:(glob)` pathspec). Deprecate modifies existing files; archive uses git renames — neither produces an 'A' diff entry. Verified end-to-end against the real commits of PRs #2267, #2268, #2269 plus `workflow_dispatch` and initial-push fallbacks — all five scenarios produce the expected behavior.

### DIFF
--- a/.github/workflows/publish-github-release.yaml
+++ b/.github/workflows/publish-github-release.yaml
@@ -115,6 +115,8 @@ jobs:
     steps:
       - name: Checkout code
         uses: actions/checkout@v6
+        with:
+          fetch-depth: 0
 
       - name: Get provider lists
         id: providers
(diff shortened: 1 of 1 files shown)


Gacko (approved):

Gacko:

Gacko:

PR #2215, opened 2026-03-01
CAPI: Release v35.0.0.

taylorbot:
**NOTE:** Keep the default apps in the provider specific sections aligned. Some apps might not be used in vSphere e.g., but in AWS and Azure and so we cannot define them globally but still have to update them for AWS and Azure.

```sh
for provider in aws eks azure vsphere cloud-director
do
  base="34.5.0"
  apps=(
    "--app" "cert-exporter@@@cilium#kyverno-crds#prometheus-operator-crd#vertical-pod-autoscaler-crd"
    "--app" "cert-manager@@@alloy-podlogs-crds#cilium#prometheus-operator-crd#vertical-pod-autoscaler-crd"
    "--app" "etcd-k8s-res-count-exporter@@@kyverno-crds#prometheus-operator-crd#vertical-pod-autoscaler-crd"
    "--app" "k8s-dns-node-cache@@@cilium#kyverno-crds#prometheus-operator-crd"
    "--app" "metrics-server@@@cilium#kyverno-crds#vertical-pod-autoscaler-crd"
    "--app" "net-exporter@@@cilium#prometheus-operator-crd#vertical-pod-autoscaler-crd"
    "--app" "node-exporter@@@kyverno-crds#prometheus-operator-crd#vertical-pod-autoscaler-crd"
    "--app" "rbac-bootstrap@0.3.0"
    "--app" "vertical-pod-autoscaler@@@prometheus-operator-crd#vertical-pod-autoscaler-crd"
  )
  drops=()

  case "${provider}" in
    aws)
      apps+=(
        "--app" "external-dns@@@cilium#prometheus-operator-crd#vertical-pod-autoscaler-crd"
        "--app" "k8s-audit-metrics@@@cilium#kyverno-crds#prometheus-operator-crd#vertical-pod-autoscaler-crd"
        "--app" "prometheus-blackbox-exporter@@@prometheus-operator-crd#vertical-pod-autoscaler-crd"
      )

      apps+=(
        "--app" "aws-ebs-csi-driver@@@kyverno-crds#vertical-pod-autoscaler-crd"
        "--app" "aws-pod-identity-webhook@@@cert-manager#prometheus-operator-crd#vertical-pod-autoscaler-crd"
        "--app" "cluster-autoscaler@@@kyverno-crds#prometheus-operator-crd#vertical-pod-autoscaler-crd"
        "--app" "cluster-autoscaler-crossplane-resources@1.0.0"
        "--app" "external-dns-crossplane-resources@0.4.0"
[...]

PR #2288, opened 2026-05-06
Fix /rebase aborting on chained conflicts across commits

njuettner:
When `/rebase` resolved conflicts in the first commit and `git rebase --continue` then paused again on a new conflict in a later commit, the workflow aborted instead of letting the resolver run on the new conflicts.

`git rebase --continue` exits non-zero whenever it pauses on a new conflict, so the previous `|| { git rebase --abort; exit 1; }` clause treated a normal "paused again" exit as fatal. The loop's `[ -d .git/rebase-merge ]` check is the correct termination signal — let it drive.

Added a no-progress safety check (HEAD doesn't advance between iterations) to guard against a stuck rebase looping forever.

Seen on https://github.com/giantswarm/releases/pull/2215 — workflow run https://github.com/giantswarm/releases/actions/runs/25443487881.

### DIFF
--- a/.github/workflows/rebase.yaml
+++ b/.github/workflows/rebase.yaml
@@ -109,6 +109,13 @@ jobs:
           # Loop: try to auto-resolve each conflict stop with the resolver.
           # If the resolver exits non-zero, it means there are conflicts in
           # files it doesn't handle — abort the rebase and fail.
+          #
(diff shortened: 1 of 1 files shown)


Gacko (approved):

njuettner:

PR #2312, opened 2026-05-21
fix(rebase-resolver): 3-way merge releases.json and kustomization.yaml

njuettner:
The /rebase comment workflow silently re-introduced releases that
master had archived, because the resolver only compared ours (new
base) and theirs (PR commit) — deletions in master were invisible
without reading the common ancestor.

It also dropped the top-level `sourceUrl`, `changelogUrl`, and
`homepage` fields from `<provider>/releases.json` on every conflict
resolution, because `releasesFile` didn't declare them.

Changes:
- Read stage 1 (common ancestor) and do a real 3-way merge in both
  `mergeReleasesJSON` and `mergeKustomization`. Entries present in
  the ancestor but missing from either side are dropped.
- Declare the trailing top-level fields on `releasesFile` so they
  round-trip.
- Regression tests covering both behaviors.

### DIFF
--- a/tools/rebase-resolver/main.go
+++ b/tools/rebase-resolver/main.go
@@ -2,13 +2,16 @@
 // */kustomization.yaml files left behind by `git rebase origin/master`.
 //
 // Both files are ordered lists keyed by semver. A conflict between two
-// release PRs is almost always a bookkeeping clash, not a semantic one, so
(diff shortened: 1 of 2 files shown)


njuettner:

Gacko (approved):

PR #2371, opened 2026-07-01
CAPA: Bring back v33.2.0

njuettner:
Reverts the deprecation and archival of CAPA v33.2.0 (archived in #2331).

v33.2.0 is still needed as an intermediate upgrade step (etcd patch) before moving to 34.x, and should not have been deprecated.


njuettner:
> 
image

Is expected to fail

yulianedyalkova (approved):

PR #2388, opened 2026-07-16
Fix release-<provider> chart versioning: publish clean semver instead of dev tags

njuettner:
Addresses: https://gigantic.slack.com/archives/C07KSM2E51A/p1783516542422869

The auto-retag pipeline stopped publishing clean semver OCI tags for
release-<provider> charts after the architect orb v8→v9 bump (#2327, merged
2026-06-01). The v9 architect/push-to-app-catalog job overrides the chart
version with `gitsemver get`, which on a branch build resolves to
`0.0.0-dev.<branch>.<timestamp>` and clobbers the release version stamped by
prepare-retag-chart. Result: releases published since June only have
`0.0.0-dev.*` tags in gsoci.azurecr.io/charts/giantswarm/release-<provider>,
no X.Y.Z.

Replace the architect/push-to-app-catalog job with a push-retag-chart job that
packages with ABS (preserving the Chart.yaml version) and publishes via the
lower-level architect/push-helm command, which does not re-version. A guard
fails the job if the packaged chart is not the expected version.

Affected releases still need a re-trigger to backfill the missing clean tags.

### DIFF
--- a/.circleci/config.yml
+++ b/.circleci/config.yml
@@ -164,20 +164,13 @@ jobs:
                   \"release_version\": \"${version}\",
                   \"chart_version\": \"${CHART_VERSION}\"
                 }},
-                {\"architect/push-to-app-catalog\": {
+                {\"push-retag-chart\": {
                   \"name\": \"push-${provider}-${version}\",
                   \"context\": \"architect\",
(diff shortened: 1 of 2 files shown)


Gacko (approved):

njuettner:

PR #2390, opened 2026-07-21
Fix install-devctl: download raw binary instead of versioned tarball

njuettner:
devctl releases no longer ship `devctl-<tag>-linux-amd64.tar.gz`. Assets are now plain binaries named `devctl-linux-amd64`, so the install action failed with "no assets match the file pattern".

Download the binary directly and `chmod +x` + `mv` it into place, instead of extracting a tarball that no longer exists.


njuettner:

Gacko (approved):

PR #2398, opened 2026-07-29
Keep deprecated releases available while a smaller active release exists

njuettner:
Towards https://github.com/giantswarm/roadmap/issues/4325

Deprecated releases were archived as soon as they were no longer in use. Archiving removes them from the management clusters, so they stop working as an upgrade target — which is what broke the planned `33.2.0` upgrade for adidas.

Archiving now requires that no smaller active release exists for the provider. The lowest active release acts as an archive floor: anything above it stays available, deprecated but installable, until the lower release is archived itself. With `33.1.1` active in `capa`, `33.2.0` would no longer be archivable.

- `tools/lowest-active-release.sh` determines the floor per provider
- `tools/check-release-archive.sh` blocks moves above the floor, and blocks archiving a release that was never deprecated
- `.github/workflows/archive-releases.yaml` skips floor-protected releases before the expensive per-release validation
- usage lookback widened from 5 minutes to 7 days, so a release is not treated as unused because its metrics were briefly missing
- both automated PR bodies now state that deprecation does not affect upgrades

Deprecation itself needed no change: it only flips `spec.state` and keeps the release in `kustomization.yaml` and `releases.json`, so the Release CR stays on the management clusters. The existing suite already asserts that every non-archived release — active or deprecated — is registered in `kustomization.yaml`, which locks that in.

Deprecated releases now accumulate in the provider directory until the floor rises, since the floor drains from the bottom one release at a time.

### DIFF
--- a/.github/workflows/archive-releases.yaml
+++ b/.github/workflows/archive-releases.yaml
@@ -66,6 +66,17 @@ jobs:
 
             echo "INFO: Processing provider directory: $PROVIDER_ROOT_DIR"
 
+            # Releases above the lowest active release must stay available even when
+            # deprecated, because clusters on that lower release may still need them as
+            # an intermediate upgrade step.
+            # See https://github.com/giantswarm/roadmap/issues/4325
+            ARCHIVE_FLOOR=$(./tools/lowest-active-release.sh "$PROVIDER_ROOT_DIR")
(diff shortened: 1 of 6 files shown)


Gacko (approved):

njuettner:

PR #2399, opened 2026-07-30
Update requests.yaml cilium-servicemonitors for AKS

T-Kukawka:
https://github.com/giantswarm/giantswarm/issues/37208

### DIFF
--- a/aks/requests.yaml
+++ b/aks/requests.yaml
@@ -1,1 +1,5 @@
-releases: []
+releases: 
+- name: ">= 35.0.0"
+  requests:
+  - name: cilium-servicemonitors
+    version: ">= 0.2.0"
--- a/aks/requests.yaml
+++ b/aks/requests.yaml
@@ -1,5 +1,5 @@
 releases: 
-- name: ">= 35.0.0"
+- name: "> 35.0.0"
   requests:
   - name: cilium-servicemonitors
     version: ">= 0.2.0"


Gacko (commented):
If it's only bumping, then there's no need to request. 🙂

T-Kukawka:
@Gacko did this just so we do not forget with next release 😅

Gacko:
I get your point, no worries. It's just both automation and `devctl` already bump everything as long as you don't opt-out and the `requests.yaml` has grown to a point where it was no longer understandable back in Vintage. In the end it's on you, just wanted to let you know that there's no real need to add a "bump to at least that version" constraint, as it only makes it more complex.

For restrictions as in "do not bump beyond this" or "have a version between A and B" it absolutely makes sense on the other hand.

Gacko (approved):

Gacko (changes requested):
One technical detail and also a reason not to add bump-only constraints:

AKS v35.0.1 would require Cilium ServiceMonitors v0.2.0 with that constraint, but it's only a patch release, so a minor bump (and since it's a v0.x.x version potentially even a breaking change) would not be allowed.

Additionally we do minor CAPI releases for node rolling changes, so even just patch of the node image (Kubernetes, Flatcar, OS tooling) would require a minor release, but since we would not bump Cilium ServiceMonitors, the CI would fail then.

We already had the latter in the past and then had to modify the `requests.yaml` to get the security patch merged.

Gacko:
Closing this as AKS v35.1.0 ships Cilium ServiceMonitors v0.2.0. 🙂

PR #2400, opened 2026-08-04
Require full E2E suite coverage on release PRs

njuettner:

`releases-test-suites` picks suites by release stage, so a PR in `stage/development` only runs `standard`, `upgrade` and `upgrade-major`. Releases were merged from that stage, so the variant suites never ran — v34.5.0 (#2391) merged without `capa/private`, `capa/china`, `capa/cilium-eni-mode`, `capz/private`, `capv/on-capa` and `capv/on-capz`.

Adds an `E2E Coverage` check that requires the full suite set for every **newly added** release in the PR, whatever stage it is in, and lists the outstanding suites with a ready-to-paste trigger comment. Deprecate/archive PRs are unaffected (they modify or rename `release.yaml`, they do not add one).

Coverage is tied to release content rather than the head SHA: a suite that passed on an earlier commit still counts until a new release's `release.yaml` changes, so README/announcement edits do not force re-runs while component bumps do. Individual suites can be waived with `/waive-suite <provider>/<suite> <reason>`; waivers are recorded in the check output and void once the release content changes.

Merge this before giantswarm/pr-gatekeeper#307, which makes the check required.

### DIFF
--- a/.github/pr-body-templates/consolidated-release-instructions.md
+++ b/.github/pr-body-templates/consolidated-release-instructions.md
@@ -104,6 +104,16 @@ For conformance tests:
 
 `/run conformance-tests PROVIDER=capa RELEASE_VERSION=29.1.0`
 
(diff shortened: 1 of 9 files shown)


Gacko (approved):

njuettner:

njuettner:

njuettner:

PR #2402, opened 2026-08-04
Rename the coverage job so it isn't mistaken for the gate

njuettner:

The E2E Coverage workflow shows two rows in the Checks tab and they read like duplicates:

- `E2E Coverage / Evaluate E2E coverage` — the job that runs the script
- `E2E Coverage` — the check run that actually gates the merge, carrying `12/21 expected suites covered · stage/development`

Renames the job to `Publish coverage report` so it is obviously subordinate to the gate, and documents both rows in `docs/workflows-e2e-coverage.md` — including why the gate row often appears grouped under an unrelated workflow name (`gitleaks / E2E Coverage`), which is check-suite labelling and cosmetic only.

### DIFF
--- a/.github/workflows/e2e-coverage.yaml
+++ b/.github/workflows/e2e-coverage.yaml
@@ -47,7 +47,10 @@ concurrency:
 
 jobs:
   coverage:
-    name: Evaluate E2E coverage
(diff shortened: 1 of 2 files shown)


njuettner:

Gacko (approved):

PR #2403, opened 2026-08-05
Publish coverage as a commit status so it is findable

njuettner:

The `E2E Coverage` check run shows up nested under an unrelated workflow — `gitleaks / E2E Coverage` — so the merge gate is buried in the Checks tab. GitHub adopts API-created check runs into an existing check suite for the same app and commit and labels the group after the first workflow that ran in it, so where it lands is not ours to choose.

Publishes the same result as a **commit status** as well. A status is always its own top-level row, next to the `ci/circleci: ...` rows:

```
● E2E Coverage    0/21 expected suites covered · stage/development
```

States: `pending` while suites are missing, `success` when complete, `failure` when an expected suite failed. A status carries only a 140-character description, so its `target_url` links to the check run with the full per-provider report, which stays as-is for `pr-gatekeeper`.

Follow-up worth doing: adding the `E2E Coverage` status to the required status checks on `master` would make coverage enforceable independently of Heimdall — and therefore not bypassable with `/skip-ci`.

### DIFF
--- a/.github/scripts/e2e-coverage.js
+++ b/.github/scripts/e2e-coverage.js
@@ -169,19 +169,43 @@ const main = async () => {
   const headSha = pr.head.sha
   log(`Evaluating coverage for PR #${prNumber} at ${headSha}`)
 
-  // Always update the existing check run instead of creating another one with the
-  // same name: pr-gatekeeper refuses to evaluate a name that matches several runs.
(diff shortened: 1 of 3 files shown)


njuettner:

Gacko (approved):

PR #2404, opened 2026-08-05
Don't let release metadata changes invalidate coverage

njuettner:

Two fields in a `release.yaml` describe the release rather than what gets installed, so changing them cannot alter a test result — but the invalidation rule treated any diff as a content change and discarded every passing suite:

- **`spec.date`** — `devctl` rewrites it on every generation ([`pkg/release/create.go:313`](https://github.com/giantswarm/devctl/blob/main/pkg/release/create.go)), so even an `/update-release` that picks up no new versions leaves a diff:

  ```diff
  -  date: "2026-08-01T09:17:35Z"
  +  date: "2026-08-04T18:22:11Z"
  ```

- **`spec.state`** — `active`, `preview` or `deprecated`. Lifecycle metadata; a release changing state does not need re-testing.

Coverage now inspects the diff instead of merely detecting one. If the only changed lines in a `release.yaml` are those fields, the content counts as unchanged and earlier suite results keep counting. Component and app versions still invalidate, and a missing patch (diff too large to inline) is treated as changed.

Tested against synthetic diffs — date-only, state-only, and both together keep results; a version change discards them; a missing patch discards them — plus the live PRs, whose tallies are unchanged (v35 `0/21`, #2391 `12/18`, archive PR "No new releases added").

### DIFF
--- a/.github/scripts/e2e-coverage.js
+++ b/.github/scripts/e2e-coverage.js
@@ -243,25 +243,30 @@ const main = async () => {
   // count, so README/announcement-only commits don't invalidate coverage.
   const commits = await paginate(`/pulls/${prNumber}/commits`)
 
-  // devctl sets spec.date to the current time on every generation (see
-  // devctl pkg/release/create.go), so even an /update-release that picks up no new
-  // versions leaves a diff. A timestamp is not something the tests can observe, so a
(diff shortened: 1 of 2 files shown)


Gacko (approved):

njuettner:

PR #2405, opened 2026-08-05
Match suite results to release content by fingerprint

njuettner:

Release branches get force-pushed routinely, and suite check runs stay attached to the commit they ran on. Coverage only looked at commits currently in the PR, so a force-push discarded everything.

Live example on #2215: the check reported `0/21` while eight suites had actually run — two passing, six failing — against release files that are **byte-identical** to the current head:

```
capa: IDENTICAL (2cd86d5f865a)      vsphere: IDENTICAL (809039614031)
azure: IDENTICAL (a19b72745638)     cloud-director: IDENTICAL (a70868d454a6)
eks: IDENTICAL (8f50c8d8068e)
```

With this change the same PR reports `2/21` and goes red on the six failures — the truthful answer, instead of a blank slate that demands a full 21-suite re-run.

**How** — results are matched by fingerprint rather than by diff: the tracked `release.yaml` files are fetched at a commit, `date` and `state` lines stripped, the rest hashed. Candidate commits are the PR's own plus the heads recovered from the timeline's `head_ref_force_pushed` events. This replaces the diff-based rule from #2404, which cannot work across a force-push — the compare API reports diverged commits against their merge base, so every release file looks rewritten.

**Request budget** — this runs on every suite completion, so: a commit is only compared when it carries a result still missing, file comparison stops at the first difference, scanning stops at the first commit whose content has moved on, and at most 25 commits are examined (logged when reached). Evaluating #2215 costs 29 requests.

Tested against the live PRs (#2215 `2/21` failure, #2391 `12/18`, archive PR "No new releases added") and synthetically: metadata-only differences count, version changes do not, a force-pushed-away commit is recovered, and a missing release file at an old commit is ignored.

### DIFF
--- a/.github/scripts/e2e-coverage.js
+++ b/.github/scripts/e2e-coverage.js
@@ -17,6 +17,7 @@
 //   CAPI_NAMES         - JSON map of provider directory → CAPI name
 //   PR_NUMBER          - PR to evaluate (workflow_dispatch only)
 
+const crypto = require('crypto')
 const fs = require('fs')
 
 const API_ROOT = 'https://api.github.com'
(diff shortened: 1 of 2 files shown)


njuettner:

Gacko (approved):

